PR #23, opened 2022-01-21
DGJ-6 Added initial OpenShift build and deploy configs

simensma-fresh:
**Openshift Templates**
This PR includes OpenShift templates required to get [formsflow.ai](https://formsflow.ai) up and running.
The templates are based on the ones found in the [formsflow.ai repo](https://github.com/AOT-Technologies/forms-flow-ai/tree/master/deployment/openshift). Note: This PR only includes the templates we made changes to.

**Web App Dockerfile**
This PR also includes a `Dockerfile` with some small changes to make it suitable for us:
- Upgraded to node 14 (to make the app run properly)
- The `Dockerfile` clones the `formsflow.ai` repo, and copies any other files from this repo on top of it so we can make changes to the webapp without having to copy over the full codebase


forms-flow-web/Dockerfile
```diff
@@ -0,0 +1,36 @@
+# base image
+FROM node:12.4.0-alpine as build-stage
```
iman-jamali-fw, line 2: If there is no problem with dependecies I suggest using the version of node LTS that is currently in ACTIVE state which is v16. 

v12 maintenance support will finish in April and v14 support for security update will finish in 1 year and 3 months.

https://endoflife.date/nodejs

forms-flow-web/Dockerfile.prod
```diff
@@ -0,0 +1,32 @@
+# Used as a reference for any custom logic to be done.
+# if the application is built from a source repo other than aot , use the custom copying.
+# steps : checkout the aot source code and use the copy logic.
+# -----------------------------------------------------------
+# base image
+FROM artifacts.developer.gov.bc.ca/docker-remote/node:14.18.3-alpine as build-stage
```
iman-jamali-fw, line 6: Same thing with here with node version. If no issue with dependencies, v16 is more preferred.

iman-jamali-fw (approved):
Great work, I mostly tried to read and figure out what openapi tempates are :)

Just a comment regarding the node version, apart from that, everything look great!

forms-flow-web/Dockerfile
simensma-fresh: Unfortunately moving to node 16 introduces some dependency issues, so think we'll have to stick with `v14`

PR #59, opened 2022-02-02
DGJ-11 Added scripts to run the project locally

simensma-fresh:
## Summary
Added config and scripts to run the project locally using `docker-compose`

## Changes
**docker-compose**
- Added docker-compose and a Makefile to run the project locally based on the ones found in https://github.com/AOT-Technologies/forms-flow-ai

**Camunda Update**
- Added custom Dockerfile / pom.xml with support for the newest Keycloak version (upgraded oauth2-oidc-sdk dependency)


.gitignore
```diff
@@ -0,0 +1,201 @@
...
+
+*.sublime*
+
+psd
+thumb
+sketch
+
+# Databases
+postgres
+mongodb
```
iman-jamali-fw, line 201: We can also add [vscode gitignore](https://github.com/github/gitignore/blob/main/Global/VisualStudioCode.gitignore)

kkapoor-fresh (commented):
Just to confirm again, we won't be setting up analytics in local

iman-jamali-fw (approved):
I think I see more docker/config files in this pr than all I've ever seen combined  :)
Looks great!

simensma-fresh:
> Just to confirm again, we won't be setting up analytics in local

I think for now we'll skip analytics locally as the project requires a lot of resources when running locally already 😰  If needed we can add it in

kkapoor-fresh (approved):

PR #69, opened 2022-02-14
Dgj56-Notify Supervisor

matbusby-fw:
## Summary

Modified the telework camunda flow to notify both the submitting user and their supervisor upon submission of the form.

## Changes

- Overrode `FormBPMDataPipelineListener.java` with custom version that filters out strings longer than 4000 characters (the base64 string of the submitter signature is often too long and would cause the listener to fail)
- Added an activity to `telework-form.bpmn` for sending a confirmation email to a supervisor.
- Rebranched the rest of the workflow off of this email, rather than the confirmation to the user.
- Created custom java class `TaskAssignmentListener.java`.  This facilitates the sending of an email as part of a task listener where the user in Camunda can specify the recipient, body and subject.
- Updated .env and yaml files with `WEB_BASE_URL`
- Added documentation to the readme in the bpm folder.  This could be moved eventually if we need.

image
## Notes
Variables with a string length longer than 3999 characters will not be available in Camunda.  I don't think this will present an issue, but it's worth noting.


forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/FormBPMDataPipelineListener.java
```diff
@@ -0,0 +1,53 @@
...
+        }
+    }
+
+    private void syncFormVariables(DelegateExecution execution) throws IOException {
+        Map<String,Object> dataMap = formSubmissionService.retrieveFormValues(String.valueOf(execution.getVariables().get("formUrl")));
+        for (Map.Entry<String, Object> entry: dataMap.entrySet()) {
+            String fieldValue = String.valueOf(entry.getValue());
+            if (fieldValue.length() < 4000) {
+                execution.setVariable(entry.getKey(), entry.getValue());
+            }
```
iman-jamali-fw, line 50: We need to add this limitation to the document. I suggest if it's not difficult or in another ticket, let the user know of this limitation on UI (add a label or something).

forms-flow-web/Dockerfile
```diff
@@ -27,7 +27,7 @@ COPY ./${CUSTOM_SRC_DIR}/  /app/${CUSTOM_SRC_DIR}/
 ARG CUSTOM_SVG_DIR=public/webfonts
 COPY ./${CUSTOM_SVG_DIR}/ /app/${CUSTOM_SVG_DIR}/
 
-COPY ./package*.json .
+COPY ./package*.json ./
```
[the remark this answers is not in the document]
matbusby-fw: This failed on your mac?  I initially tested it on linux, but then I tried it with the `/` on my mac and everything installed fine for me.  Interesting.  Any thoughts here @simensma-fresh?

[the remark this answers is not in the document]
simensma-fresh: I guess if both versions cause issues, we could test out using the full path of the folder:

```
COPY ./package*.json /app
```

forms-flow-bpm/src/main/resources/processes/telework-form.bpmn
```diff
@@ -73,40 +74,77 @@ task.execution.setVariable('deleteReason', "completed");</camunda:script>
...
+          <camunda:inputOutput>
+            <camunda:inputParameter name="subject">FOR ACTION: A new Telework Agreement has been submitted</camunda:inputParameter>
+            <camunda:inputParameter name="to">${managerEmployerEMail}</camunda:inputParameter>
+            <camunda:inputParameter name="text">Hello
+
+
+For your review, please find attached a new Telework Agreement submitted by ${submitterName}.
+
+
+Go to the Manager/Supervisor Sign-Off Section at ${formUrl} to complete and submit this form.
```
simensma-fresh, line 95: I think we might have to change this a bit, ideally the link would take you to the url of the task associated with it, and not the form itself 😅

simensma-fresh (changes requested):

forms-flow-bpm/src/main/resources/processes/telework-form.bpmn
```diff
@@ -73,40 +74,77 @@ task.execution.setVariable('deleteReason', "completed");</camunda:script>
...
+      <bpmn:extensionElements>
+        <camunda:executionListener class="org.camunda.bpm.extension.hooks.listeners.FormBPMDataPipelineListener" event="take" />
+      </bpmn:extensionElements>
+    </bpmn:sequenceFlow>
+    <bpmn:serviceTask id="Activity_13ovu4t" name="Send Approval Email to Supervisor">
+      <bpmn:extensionElements>
+        <camunda:connector>
+          <camunda:inputOutput>
+            <camunda:inputParameter name="subject">FOR ACTION: A new Telework Agreement has been submitted</camunda:inputParameter>
+            <camunda:inputParameter name="to">${managerEmployerEMail}</camunda:inputParameter>
```
simensma-fresh, line 88: Thoughts on updating this to `managerEmail`? Or something a bit simpler as we'll have to make sure it aligns with the auto-population work
CC/ @iman-jamali-fw

forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/FormBPMDataPipelineListener.java
iman-jamali-fw: Probably the one @simensma-fresh mentioned on how to use Camunda. Maybe in the Camunda readme?

forms-flow-bpm/src/main/resources/processes/telework-form.bpmn
matbusby-fw: Oh yeah!  That makes sense.  This is proving to need a fair bit of rework.  The way we're currently sending the email, the task hasn't been created at that point, so we have no way of including the task id for reference.  Have to write a new java class listener to send an email when the user task is created.

forms-flow-bpm/src/main/resources/processes/telework-form.bpmn
matbusby-fw: Sure, that works.  I just used what was in place on the current iteration of the telework form on dev, but that does make sense.

simensma-fresh (approved):

iman-jamali-fw (approved):
LGTM!

Issue #11, opened 2022-01-13
Branded look-and-feel for forms


schuyt:
As a digital journey user I would like to see a branded consistent look-and-feel for forms

**Acceptance Criteria**
- Styles from the BC Gov Design System:  https://developer.gov.bc.ca/Design-System/Navigation-Bar-Basic are applied to the listed components
- Import and apply government approved BC Sans font as per the BC Gov Design System
- Consider Colours match style guide
- Consider font-size & colours for accessibility
- Page header is updated to match the styling provided by the BC Gov Design System [navigation-bar-basic](https://developer.gov.bc.ca/Design-System/Navigation-Bar-Basic)
- Page footer is updated to match the one provided by the BC Gov Design System [Footer-Basic](https://developer.gov.bc.ca/Design-System/Navigation-Bar-Basic)
- The header of the Telework Agreement is updated to match the given form [Logo](https://drive.google.com/file/d/1E6FxH_t8BcXM5f_6aVity5NpJnAhWFla/view?usp=sharing). [https://www.figma.com/file/gQDAaN0lnJJj1xg6NXFEsI/Wireframes?node-id=11%3A194](https://www.figma.com/file/gQDAaN0lnJJj1xg6NXFEsI/Wireframes?node-id=11%3A194)

**References**:
- Button inc has a react library with alot of prebuilt BCGov branded components: https://button-inc.github.io/service-development-toolkit/@button-inc/bcgov-theme/index.html?path=/docs/components-navigation--html  https://github.com/button-inc/service-development-toolkit

- [Visual Identity Guidelines](https://www2.gov.bc.ca/gov/content/governments/services-for-government/policies-procedures/bc-visual-identity)
- [Design System](https://developer.gov.bc.ca/Design-System/About-the-Design-System)
- [Example](https://www2.gov.bc.ca/assets/gov/careers/forms-tools/managers-supervisors/c19_vax_exempt_medical.pdf) of branded form
- [Colour Pallet](https://developer.gov.bc.ca/Colour-Palette)
- [Contrast Accessibility Checker Tool](https://webaim.org/resources/contrastchecker/)
- 

**Notes**:
- Will use Corporate identity "Where We Work" for these forms - see mock-up
- Review styles that we have afterwards and capture any required changes
- Spin out any stories for adjustments

schuyt:
[...]

PR #83, opened 2022-02-18
Fix/dgj 11 branding

iman-jamali-fw:
## Summary

Fixing the style issues with branding ticket #11 

## Changes
fixing include
- header
- navigation items, sign out button on small screen
- footer links
- buttons style fixed

More details in this [doc](https://docs.google.com/document/d/1Hjbg-NDL3WRvXPXw8PqDPV9eEvSLO_EEasi0-r6mfrQ/edit)

### DIFF
--- a/forms-flow-web/src/components/Form/List.scss
+++ b/forms-flow-web/src/components/Form/List.scss
@@ -1,0 +1,94 @@
+.grid-forms
+{
+    .col-md-4 {
+        flex: 0 0 50% !important;
+        max-width: 50% !important;
+        padding-top: 15px;
+    }
+    .col-md-8 {
+        flex: 0 0 50% !important;
+        max-width: 50% !important;
+    }
+}
+.custom-grid{
+    .col
+    {
+        border-left: 1px solid #ced4da;
+        min-height: 50px;
+        text-align: start;
+        padding-top: 15px;
+        height: auto;
+        word-break: break-all;
+    }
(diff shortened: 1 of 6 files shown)


forms-flow-web/src/components/Form/List.scss
```diff
@@ -0,0 +1,94 @@
...
+
+.flex-item-right {
+  justify-content: flex-end;
+  display: flex;
+  flex-direction: row;
+  flex: 50%;
+}
+.btn-left{
+    margin: 10px;
+    height: 50px;
```
iman-jamali-fw, line 87: only this `height: 50px;` property changed to fix the issue with button style on this page

simensma-fresh (approved):
🎨

matbusby-fw (approved):
Nice work!  It's odd reviewing some of this code.  I noticed a few things that were from the original formsflow code (not written by you) that might merit a note, but I don't think we want to go down the rabbit hole of patching code we're not writing. (example of this was not using a key on the mapped list items)

iman-jamali-fw:
> Nice work! It's odd reviewing some of this code. I noticed a few things that were from the original formsflow code (not written by you) that might merit a note, but I don't think we want to go down the rabbit hole of patching code we're not writing. (example of this was not using a key on the mapped list items)

Agreed! certainly style of the app is a little confusing mixing form.io style with form-flow and our own :)

PR #84, opened 2022-02-18
DGJ58-Calculation of Telework Days

matbusby-fw:
## Summary

Added custom class `alert-info` to the styles.css.  This can be applied to a form element in its `Custom CSS Class` section.
Also added documentation with an example of how to create calculated values for fields.  This example shows the code necessary to calculate the number of days teleworking base on the boxes checked.  This example code will need to be added to a form directly so is not representing in the base code other than the doc

## Changes

- Added `alert-info` to styles.css
- Added docs folder and `forms.md` within it.
- Add some documentation to `forms.md`

image


forms-flow-web/src/styles.scss
```diff
@@ -681,3 +681,31 @@ a:hover {
...
+  background-color: #F9F1C6;
+  padding: 15px;
+  display: flex;
+  justify-content: space-evenly;
+  align-items: center;
+  &:before {
+    margin-right: 11px;
+    content: url(./assets/warning.svg);
+  }
+;
```
iman-jamali-fw, line 695:
```suggestion
  };
```

simensma-fresh (approved):

PR #91, opened 2022-02-22
Fix/dgj 20 auto populate user data

iman-jamali-fw:
## Summary

Populate default values improved

## Changes
BE
  - Home address added 
  - Check for empty GUID on API was improved
  
FE
  - State for initial user data loading added
  - Loading UI overlay on form until employee data is loaded
  - Form can use the same employee data field multiple times. This is possible by adding the field name + "_" + "arbitrary name".
Screen Shot 2022-02-22 at 8 24 35 AM
Screen Shot 2022-02-22 at 8 25 22 AM

## Notes
The way using the same employee data field multiple times works is that when the form is building, the user should add "_" + "another name" to a field if it was used once.

For example, if you want to use the "first name" of an employee in more than one field in a form you should:
first filed name can be: `firstName`
Second field name: `firstName_xxx`
Third field: `firstName_yyy`

you get the idea

### DIFF
--- a/forms-flow-api/src/formsflow_api/models/employee_data.py
+++ b/forms-flow-api/src/formsflow_api/models/employee_data.py
@@ -3,6 +3,8 @@ class EmployeeData():
    self.firstName = data["first_name"]
    self.lastName = data["last_name"]
    self.email = data["email"]
+   self.address1 = data["address1"]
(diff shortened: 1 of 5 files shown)


simensma-fresh (approved):
LGTM!

PR #98, opened 2022-02-23
using user data multiple time in a form improved to the forms with ne…

iman-jamali-fw:
## Summary
My earlier solution to use user data multiple times for auto-populate in a form didn't work in the forms with nested keys. 

## Changes
* A function was used to search for form field keys recursively regardless of how the form is designed. This will make the solution more generic and let this feature be used in many forms.
* A new employee user data (name) was added as it is been used in Telework form.


matbusby-fw (approved):
Nice!

simensma-fresh (approved):

PR #102, opened 2022-02-23
auto-populate document was added

iman-jamali-fw:
## Summary

Added documentation for how to use the auto-populate feature on top of what @matbusby-fw initially added.


docs/forms.md
```diff
@@ -32,3 +33,12 @@ return acc;
...
 ```
 
+## How to auto-populate form fields with the user data
+
+If you would like to automatically populate a form's field with the user's data, you can do so by adding one of the pre-defined names to a field in your form. Follow these steps to add the user's data to a field:
+
+Step 1: Look at the pre-defined values along with their names in `/forms-flow-api/src/formsflow_api/models/employee_data.py` file. These values are available to all forms.<br>
+Step 2: When creating forms, in the Form builder dialog for a component, under `API` tab, you can set the `Property Name` of a field. If you want this field to be pre-populated with the user data simply add one of the pre-defined names (found in the last step) as a `Property Name`.<br>
+Step 3: If you want to use one piece of user data more than once in a form, you can append `_` to the property name of your field. For example, if you want to use the user's `name` second time in a form, you can set the `Property Name` of the field to be `name_2`.
+
```
matbusby-fw, line 43: What do you think about adding a list of the available variables here?

matbusby-fw (approved):
Documentation!  Huzzah!

docs/forms.md
iman-jamali-fw: Hmm, it is not a bad idea. Two issues though, one is we have to keep it updated each time values change and also not sure about privacy of showing type of employee data publically. @simensma-fresh What do you think?

matbusby-fw: Oh yeah.  Interesting point about the privacy.  It feels like it shouldn't be an issue with just the types of information listed, but you never know with government.  

Is the list of available variables subject to a lot of change?  Having to do a few updates isn't so bad, but if it's constant it makes it less manageable.  I have a feeling they will be asking for these documented at some point for their designers, but maybe there's a better place to show these.

iman-jamali-fw: OK, now that I'm thinking, the types of info are already pubic/accessible in the code :) so technically there shouldn't be an issue putting it here.

For updates, I don't think it changes a lot.

simensma-fresh: I would suggest that we add in a simple table to list all the properties. I think most forms will be built by non-developers, so having something to view that's not code will be great for handoff purposes :) 

Even just something like

| Property  | Description            |
|-----------|------------------------|
| firstName | First Name of Employee |
| lastName  | Last Name of Employee  |

iman-jamali-fw: Addressed

simensma-fresh (approved):

PR #107, opened 2022-02-24
A small improvement for auto-populate

iman-jamali-fw:
## Summary
A small improvement for auto-populate to only populate the form when form and employee data are loaded and showing the proper loading message.

### DIFF
--- a/forms-flow-web/src/components/Form/Item/View.js
+++ b/forms-flow-web/src/components/Form/Item/View.js
@@ -62,6 +62,14 @@ const View = React.memo((props) => {
   }
 
   const getDefaultValues = (data) => {
+    if (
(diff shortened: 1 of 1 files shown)


matbusby-fw (approved):
Looks great!  This should keep the fields from being cleared after the form has been loaded for users without autopopulate info.

iman-jamali-fw:
> Looks great! This should keep the fields from being cleared after the form has been loaded for users without autopopulate info.

Actually, it doesn't! It should be dealt with in another ticket. But that was the original intent, I removed that part (partial fix) and kept the rest of the improvement code.

simensma-fresh (approved):

PR #138, opened 2022-03-08
DGJ-94 Fix upload of files after removing old one

simensma-fresh:
## Summary
Fixes an issue where you couldn't re-upload a new file after removing a file that was too big. Also fixed an issue where form submissions would time out when running the app locally (CC/ @iman-jamali-fw, think this fixes the issue you experienced), and handling a user that doesn't exist in the ODS gracefully instead of producing a 500 error. 

## Changes
- Added event listener to all `fileStatusRemove` elements that on click shows the file upload component again
- Added lifecycle badge to the repo (required for all BCGov repos)
- Increased # threads for running the WebAPI locally to avoid deadlock that happens when submitting the form
- Return `400` instead of `500` if you look up a user that doesn't existing in the ODS

image

### DIFF
--- a/forms-flow-api/src/formsflow_api/services/employeeDataService.py
+++ b/forms-flow-api/src/formsflow_api/services/employeeDataService.py
@@ -17,13 +17,12 @@ class EmployeeDataService:
         response_from_BCGov = requests.get("{}?$filter=GUID eq '{}'".format(employee_data_api_url, guid),
                        headers={"Authorization": test_auth_token})
       except:
-        return {"message": "Something went wrong!"}, HTTPStatus.INTERNAL_SERVER_ERROR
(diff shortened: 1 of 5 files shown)


forms-flow-api/src/formsflow_api/services/employeeDataService.py
```diff
@@ -17,13 +17,12 @@ def get_employee_data_from_bcgov(guid):
...
 
-      return emp_data.__dict__
-      
-
-    
\ No newline at end of file
+      if employee_data_res and employee_data_res["value"] and len(employee_data_res["value"]) > 0:
+        emp_data = EmployeeData(employee_data_res["value"][0])
+        return emp_data.__dict__, HTTPStatus
+      return {"message": "No user data found"}, HTTPStatus.NOT_FOUND
```
iman-jamali-fw, line 28: Much better error handling!

PR #140, opened 2022-03-09
DGJ22-Generate-PDF

matbusby-fw:
## Summary

Created a custom Java class listener that will email both the user and the manager a pdf copy of the form on the event it is attached to in Camunda

## Changes

- Recreates the form based on it's fields and values
- Currently uses Formio default styling
- Requires the designer to add the class to a task in Camunda, but does not require any variables to be passed in
- Added documentation in the bpm readme

image

## Notes

If we want this to be more dynamic, we can allow the passing of variables into the class so designers can add custom recipients, body text, subject etc.  For the moment however I think making those parts static for this listener makes the workflow simpler.


forms-flow-bpm/Dockerfile.prod
```diff
@@ -4,7 +4,7 @@
 # -----------------------------------------------------------
 
 # Maven build
-FROM maven:3.6.1-jdk-11-slim AS MAVEN_TOOL_CHAIN
+FROM mcr.microsoft.com/openjdk/jdk:11-ubuntu AS ubuntu-jdk
```
simensma-fresh, line 7: I think we'll need to add the maven installation snippet here as well:

```
RUN apt-get update \
  && apt-get install -y curl git \
  && rm -rf /var/lib/apt/lists/*

ARG MAVEN_VERSION=3.8.4
ARG USER_HOME_DIR="/root"
ARG SHA=a9b2d825eacf2e771ed5d6b0e01398589ac1bfa4171f36154d1b5787879605507802f699da6f7cfc80732a5282fd31b28e4cd6052338cbef0fa1358b48a5e3c8
ARG BASE_URL=https://apache.osuosl.org/maven/maven-3/${MAVEN_VERSION}/binaries

RUN mkdir -p /usr/share/maven /usr/share/maven/ref \
  && curl -fsSL -o /tmp/apache-maven.tar.gz ${BASE_URL}/apache-maven-${MAVEN_VERSION}-bin.tar.gz \
  && echo "${SHA}  /tmp/apache-maven.tar.gz" | sha512sum -c - \
  && tar -xzf /tmp/apache-maven.tar.gz -C /usr/share/maven --strip-components=1 \
  && rm -f /tmp/apache-maven.tar.gz \
  && ln -s /usr/share/maven/bin/mvn /usr/bin/mvn

ENV MAVEN_HOME /usr/share/maven
ENV MAVEN_CONFIG "$USER_HOME_DIR/.m2"

FROM ubuntu-jdk AS bpm
```

forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/FormPDFListener.java
```diff
@@ -0,0 +1,319 @@
...
+            Map<String, Object> data = new HashMap<>();
+            data.put("formFields", formFields );
+            data.put("formValues", formValues);
+
+            // Generate the template from '/templates/form.html'
+            Template template = getTemplate();
+
+            // Save the template as output.html
+            Writer fileWriter = new FileWriter(new File("output.html"));
+            template.process(data, fileWriter);
```
simensma-fresh, line 93: Ideally we would write this to a temporary file so avoid any chance of `output.html` being reused by someone else for example two requests coming in at the same time.

```java
File f = null;
try {
        f = File.createTempFile();
        template.process(data, new FileWriter(f));
        RenderPage(f);
} finally {
     if(f != null) {
       f. deleteIfExists();
     }
}      

```

forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/FormPDFListener.java
```diff
@@ -0,0 +1,319 @@
...
+    private String submitterEmail;
+
+    @Autowired
+    private FormSubmissionService formSubmissionService;
+
+    @Autowired
+    private HTTPServiceInvoker httpServiceInvoker;
+
+    @Override
+    public void notify(DelegateTask delegateTask) {
```
simensma-fresh, line 66: I'm wondering if we maybe could combine this listener with the `TaskAssignmentListener` to make it a bit more flexible and the PDF to be included in any email you desire. 

E.g. allow passing along a `attachSubmission` field that determines whether or not the PDF should be attached.

image

simensma-fresh (changes requested):
Looks great! Just a couple of smaller things and one (bigger) which would be to see if we can combine this with the existing `TaskAssignmentListener`.

For the purpose of demo tomorrow, I think we can merge this after the `Dockerfile.prod` comment has been fixed

forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/FormPDFListener.java
iman-jamali-fw: pro tip :)

iman-jamali-fw (approved):
Great great work! 
A lot of learning in this work. Even for the reviewer :)
Well done @matbusby-fw!

forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/TaskAssignmentListener.java
```diff
@@ -16,77 +25,180 @@
...
 
-            } catch (Exception e) {
-                throw new MailConnectorException("Failed to send mail: " + e.getMessage(), e);
+                } catch (Exception e) {
+                    throw new MailConnectorException("Failed to send mail: " + e.getMessage(), e);
+                }
             }
-
+        } catch (Exception e) {
+            e.printStackTrace();
```
simensma-fresh, line 111: I think we should try to handle this exception gracefully instead of just printing the exception.

A pattern used in other Listeners (for example FormBPMDataPipelineListener) would be to move all this logic into a separate method, and passing any exception along to `handleException` of the `BaseListener`.

Example:

```java
    @Override
    public void notify(DelegateTask delegateTask) {
        try {
            sendEmail(delegateTask.getExecution());
        } catch (Exception e) {
            handleException(delegateTask.getExecution(), ExceptionSource.TASK, e);
        }
    }
```

forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/TaskAssignmentListener.java
```diff
@@ -16,77 +25,180 @@
...
+        JSONObject formValues = formSubmissionService.retrieveFormJson(formUrlString).getJSONObject("data");
+
+        // Get the URL for the form fields
+        String formUrl = formUrlString.split("/submission/")[0];
+        JSONObject formFields = formSubmissionService.retrieveFormJson(formUrl);
+
+        // Extract the emails for the manager and submitter
+        Map<String, Object> valuesDataMap = formSubmissionService.retrieveFormValues(formUrlString);
+//        managerEmail = String.valueOf(valuesDataMap.get("managerEmail"));
+//        submitterEmail = String.valueOf(valuesDataMap.get("email"));
```
simensma-fresh, line 127: Can remove these

forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/TaskAssignmentListener.java
```diff
@@ -16,77 +25,180 @@
...
+
+            Browser browser = playwright.chromium().launch();
+            Page page = browser.newPage();
+
+            page.navigate(file.toPath().toUri().toURL().toString());
+            page.waitForLoadState(LoadState.DOMCONTENTLOADED);
+
+            page.pdf(new Page.PdfOptions().setPath(Paths.get("form.pdf")).setMargin(new Margin().setRight("40px").setLeft("40px").setTop("40px").setBottom("40px")));
+        } catch (Exception e) {
+            e.printStackTrace();
```
simensma-fresh, line 198: Similar to my comment above, we should try to handle this one. Maybe add in a `throws Exception` to the method signature, and let it be caught by the exception handler at the top level

forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/TaskAssignmentListener.java
```diff
@@ -252,12 +359,22 @@ private String getSubject(DelegateExecution delegateExecution) {
         return String.valueOf(this.subject.getValue(delegateExecution));
     }
 
-    private String getRecipient(DelegateExecution delegateExecution) {
-        return String.valueOf(this.recipientEmail.getValue(delegateExecution));
+    private InternetAddress[] getRecipients(DelegateExecution delegateExecution) throws AddressException {
+        String[] recipientArray = String.valueOf(this.recipientEmails.getValue(delegateExecution)).split(",");
+        InternetAddress[] recipients = new InternetAddress[recipientArray.length];
```
simensma-fresh, line 364: Can make this a bit cleaner by using a Java stream + map instead of using a for loop. Maybe something like this:

```java
InternetAddress[] recipients = Arrays.stream(recipientArray)
       .map(r -> InternetAddress.parse(r)[0])
       .toArray(InternetAddress[]::new)

```

simensma-fresh (changes requested):

simensma-fresh (approved):

PR #148, opened 2022-03-17
DGJ-22 Fixed cache permission issue when rendering PDF

simensma-fresh:
## Summary
This PR fixes an issue when rendering a PDF in dev using Playwright: it fails with a permission denied error in the /.cache folder.
It also fixes an issue where the ODS endpoint to send submissions to is constructed wrong 

## Changes
- Updated Dockerfile.prod to create a /.cache folder + update permissions accordingly
- Updated ODS url to use `#` instead of `/` as a separator for the endpoint part


matbusby-fw (approved):
Nice catch!

PR #151, opened 2022-03-18
DGJ-135 Delay rendering of form for PDF

simensma-fresh:
## Summary

Add a `setTimeout` to wait for the page to complete rendering before triggering PDF generation. Now the generation waits for a `#formio-form-rendered` element to be inserted into the page before starting.
Should (hopefully) fix an issue in dev where the PDF is sometimes blank.

## Changes
- Insert an empty `#formio-form-rendered` div after setting the form submission
- Wait for `#formio-form-rendered` before rendering


matbusby-fw (approved):
Ah, I see.  I removed the original `Thread.sleep` based on Iman's comment, but I guess `LoadState.DOMCONTENTLOADED` didn't do the trick on the server.  Worked locally though :man_shrugging:

simensma-fresh:
YEah it *sometimes* seems to work in dev as well, hopefully this is a bit more consistent :)

PR #171, opened 2022-03-25
DGJ-72 Split email sending for telework form into separate tasks

simensma-fresh:
## Summary
- Fixed setting of the `managerSignatureTimestamp` variable - which I believe might have caused some of the email sending issues in dev 👀 
- Moved email sending for telework form into separate tasks and added transaction boundaries to make issues show up in Camunda + not roll back any tasks that already completed on failure. Should prevent e.g. double sending of emails if you retry a failed task

## Changes
-  Modified the TaskAssignmentListener to allow usage through service tasks
- Updated Telework Form bpmn to fix email sending issues

image
image
image


iman-jamali-fw (approved):
LGTM!

PR #181, opened 2022-03-29
Fix/dgj 169 auto populate data

iman-jamali-fw:
## Summary

This pr fixes the issue with auto-populate data from ODS reported in [this ticket](https://app.zenhub.com/workspaces/digital-journeys-61c10fe84c8bcb001491c5f5/issues/bcgov/digital-journeys/169)

## Changes
- A small fix to `employeeDataService.py`.
- The flag that previously was added to stop setting the default value s(from ODS) to the form was removed.
- Auth token refresh was no longer dispatches the token. This dispatch was the root cause that made the form refresh every 5 mins (depending on the auth token refresh interval value).

## Notes
Please read more about the auth token refresh issue [here ](https://github.com/bcgov/digital-journeys/pull/149)
We may want to watch in case any issue came up after stopping auth token dispatch.

### DIFF
--- a/forms-flow-web/src/services/UserService.js
+++ b/forms-flow-web/src/services/UserService.js
@@ -73,14 +73,17 @@ const initKeycloak = (store, ...rest) => {
 let refreshInterval;
 const refreshToken = (store) => {
   refreshInterval = setInterval(() => {
-    KeycloakData && KeycloakData.updateToken(-1).then((refreshed)=> {
-      if (refreshed) {
-        localStorage.setItem('authToken', KeycloakData.token);
-      }
-    }).catch( (error)=> {
-      console.log(error);
-      userLogout();
-    });
+    KeycloakData &&
+      KeycloakData.updateToken(5)
+        .then((refreshed) => {
+          if (refreshed) {
+            localStorage.setItem("authToken", KeycloakData.token);
+          }
+        })
+        .catch((error) => {
+          console.log(error);
+          userLogout();
+        });
   }, 5000);
 }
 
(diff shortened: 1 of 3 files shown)


forms-flow-web/src/services/UserService.js
```diff
@@ -0,0 +1,157 @@
...
+        console.warn("not authenticated!");
+        doLogin();
+      }
+    });
+};
+let refreshInterval;
+const refreshToken = (store) => {
+  refreshInterval = setInterval(() => {
+    KeycloakData && KeycloakData.updateToken(-1).then((refreshed)=> {
+      if (refreshed) {
```
iman-jamali-fw, line 77: Only this part was changed

forms-flow-web/src/services/UserService.js
```diff
@@ -0,0 +1,157 @@
...
+      } else {
+        console.warn("not authenticated!");
+        doLogin();
+      }
+    });
+};
+let refreshInterval;
+const refreshToken = (store) => {
+  refreshInterval = setInterval(() => {
+    KeycloakData && KeycloakData.updateToken(-1).then((refreshed)=> {
```
simensma-fresh, line 76: If I remember correctly, this one was `updateToken(5)` instead of `-1` 🙃

simensma-fresh: Just so we refresh the token when it has < 5 min left until expiry

simensma-fresh (changes requested):

forms-flow-web/src/services/UserService.js
iman-jamali-fw: Oops! You're right. let me fix it.

simensma-fresh (approved):

PR #201, opened 2022-04-04
Fixed missing IDIR / GUID parameters when Starting a Camunda process

simensma-fresh:
## Summary
The modifications to pass along IDIR/GUID when starting a camunda process were missing, which would explain why they've been missing in dev/test and just working locally

## Changes
- Added `application.py` with IDIR/GUID modifications

### DIFF
--- a/forms-flow-api/src/formsflow_api/services/application.py
+++ b/forms-flow-api/src/formsflow_api/services/application.py
@@ -58,8 +58,6 @@ class ApplicationService:
 
                 application.update({"process_instance_id": camunda_start_task["id"]})
             except TypeError as camunda_error:
-                print('hiiiiiiiii')
-                print(camunda_error)
                 response = {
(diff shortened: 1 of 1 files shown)


forms-flow-api/src/formsflow_api/services/application.py
```diff
@@ -0,0 +1,443 @@
...
+            application.update({"process_instance_id": data["process_instance_id"]})
+        else:
+            payload = {
+                "variables": {
+                    "applicationId": {"value": application.id},
+                    "formUrl": {"value": application.form_url},
+                    "formName": {"value": application.application_name},
+                    "submitterName": {"value": application.created_by},
+                    "GUID": {"value": user_guid},
+                    "IDIR": {"value": idir},
```
simensma-fresh, line 50: This is the piece that was missing

PR #318, opened 2022-06-14
scroll bar added, css for print were added

iman-jamali-fw:
## Summary

This PR enables a scroll bar on the forms and let user print forms through the browser. 

@warrenchristian1telus is the original author of this pr, I just copied them into the global CSS and tested them on my local.

## Changes
- Add CSS rules to the website's global CSS


warrenchristian1telus (approved):
Looks good - thanks @iman-jamali-fw !

PR #380, opened 2022-06-28
Keycloak Group assignment listener was added

iman-jamali-fw:
## Summary

This PR adds a Camunda Listener that can assign a Keycloak group to a supervisor user.

## Changes
- Keycloak Admin client was added to BPM
- A Camunda listener was added to dynamically add a group to a user that is marked as a supervisor on a form

Screen Shot 2022-06-27 at 7 02 11 PM

### DIFF
--- a/forms-flow-bpm/pom-docker.xml
+++ b/forms-flow-bpm/pom-docker.xml
@@ -310,6 +310,13 @@
 			<version>5.3.4</version>
 		</dependency>
 
+		<!-- https://mvnrepository.com/artifact/org.keycloak/keycloak-admin-client -->
+		<dependency>
+				<groupId>org.keycloak</groupId>
+				<artifactId>keycloak-admin-client</artifactId>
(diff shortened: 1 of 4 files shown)


warrenchristian1telus (commented):
We will need to revisit the versions as it was flagged in Snyk, but this will be addressed in a future release. Looks good to me.

warrenchristian1telus:
@iman-jamali-fw I have re-deployed this to dev. Please verify.

Also, please note that with the latest changes to the DeploymentConfigs, to contain changes to one namespace we've removed the capability for DC's to rebuild based in the image change trigger. This means when an image is updated, the related pod(s) need to be manually killed to rebuilt to :latest. This also means that to migrate an approved change (:latest) to test or prod, once built for dev, you can just restart the test / prod pods to update (no deployment necessary). But it will always go to :latest, so please be aware. This includes anything triggered by Platform Services - cluster upgrades, repairs, etc.).

PR #381, opened 2022-06-28
bugfix/maven-version-update

david-sabadash:
updated maven build version for bpm container with currently available maven version and updated SHA


iman-jamali-fw (approved):
Good catch!

PR #382, opened 2022-06-28
…

david-sabadash:
created override navBar component to change banner icon, added missing manifest and entrypoint index.html to allow changing the app name


iman-jamali-fw (approved):
Looks good to me

PR #433, opened 2022-07-19
Success Page

david-sabadash:
## Summary
created a success page for submitting and reviewing a form, form body renders conditionally based on which method was used to trigger the redirect

## Changes

Added a new route and page to the app for the success component


iman-jamali-fw (approved):
This looks great!

simensma-fresh (approved):
👍 
Looks good! Longer term we'll have to make this configurable per form, but think this will do for now

PR #435, opened 2022-07-20
DGJ-157 PDF Generation Service

simensma-fresh:
## Summary
Fixed the issue where PDF styling was off, signature blocks did not show up, and that some lines were cut off.

How?
- Moved generation of PDF over to the file upload service (exposes `GET file/pdf` to generate a pdf for the given submission). This also has the benefit that we can remove the Playwright dependency from Camunda - to avoid PDF generation potentially bringing the service down if something goes wrong.
- Switched to Puppeteer + added some `style` tweaks to fix the page break issues.

*Other Changes*


formio-upload/pdf-generation/generatePdf.js
```diff
@@ -0,0 +1,105 @@
...
+ * Local styles to embed onto the page.
+ * These are added by Puppeteer to avoid extra network requests
+ */
+const externalStyles = [
+  'pdf-generation/styles/bootstrap-4.1.3.min.css',
+  'pdf-generation/styles/font-awesome-4.7.0.min.css',
+  'pdf-generation/styles/formio-4.13.1.full.min.css',
+  'pdf-generation/styles/semantic-2.4.1.min.css',
+  'pdf-generation/styles/digital-journeys-0-1-0-min.css',
+]
```
iman-jamali-fw, line 27: Nice to have these scripts added here instead of the HTML!

formio-upload/pdf-generation/generatePdf.js
```diff
@@ -0,0 +1,105 @@
...
+        printBackground: true,
+        margin: {
+          top: '60px',
+          right: '0px',
+          bottom: '64px',
+          left: '0px'
+      },
+    });
+    
+    return pdf.toString('binary');
```
iman-jamali-fw, line 83: This is great! returning the binary is much better/cleaner than what I did (save pdf as a file and plan to send it back).

iman-jamali-fw (approved):
This is a great job! I really liked your approach of getting rid of unnecessary dependencies and generating the pdf purely with Puppeteer.

david-sabadash (approved):

PR #436, opened 2022-07-20
Implement image tags for web and webapi

kkapoor-fresh:
## Summary

Enable image tags for builds for each environments.  Currently, image builds are getting created with a "latest" tag for all environments. This change will add env specfic tags to each build which will ensure that only required deployments get  refreshed/recreated in the appropriate environment.

## Changes
- Changed deploy.yaml to pass appropriate image tags to the CI/CD pipeline
- Changed build config and deployment config templates for web and api to incorporate relevant image tags

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -42,7 +42,7 @@ jobs:
       BUILD_ID: ${{ github.event.number }}
       BUILD_NAMESPACE: d89793-tools
       BRANCH: main
-      BUILD_TAG: latest
+      BUILD_TAG: ${{ github.event.inputs.environment }}
       APP: forms-flow-web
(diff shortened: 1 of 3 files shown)


simensma-fresh (changes requested):
Looks good. We'll have to also do the same for the other services that are not managed by the GH Actions.
- Formio
- Redash
- Camunda
- File service

simensma-fresh:
> Looks good. We'll have to also do the same for the other services that are not managed by the GH Actions.
> 
> * Formio
> * Redash
> * Camunda
> * File service

Confirmed with @kkapoor-fresh That there will be a separate PR with the other services once these have been verified working 🙃

warrenchristian1telus (approved):
Looks good.

PR #440, opened 2022-07-20
DGJ 336 Logo update

david-sabadash:
## Summary
removed app title changes, updated logo changes. TEST PR - This might break the deployment, but I think the source of the issue was coming from either the index.html or manifest.json changes

## Changes

Removed DGJ 164 changes, fixed navbar import


simensma-fresh (approved):
Let's give it a try 🙃

PR #451, opened 2022-07-22
DGJ 164 App title and manifest

david-sabadash:
added index.html and manifest.json to control the application title and metadata

## Changes

added index.html and manifest.json


david-sabadash:
This shouldn't cause the environments to break if its actually the Navigation override that is the culprit

simensma-fresh (approved):

iman-jamali-fw (approved):

PR #458, opened 2022-07-26
DGJ-416 Enable cleanup of telework form submission history

simensma-fresh:
## Summary
Remove finished Telework form processes from after submission. Doing this by setting "time to live" to 0 days, which means that it will be cleaned out of the DB during the first history Cleanup Batch Window after completion.

## Changes
- Added history ttl to telework form bpmn
- Added historyCleanupBatchWindow (daily, 1am - 5am)


iman-jamali-fw (approved):
Looks great!
We just need to make sure, this doesn't conflict with the time of deleting incomplete submissions after 30 days. 

We can test this once all relevant prs are deployed to DEV.

david-sabadash (approved):

simensma-fresh:
Yup! The historyTimeToLive property only apply to processes that have been completed, so does not conflict with the 30day change

PR #460, opened 2022-07-26
DGJ-459 Use manager role to determine task tab visibility

simensma-fresh:
## Summary
- Use the `manager` role to determine whether or not tasks tab should show up
- Grant `formsflow/manager` group assignment to manager when telework form has been submitted


iman-jamali-fw (approved):
Looks good to me!

PR #469, opened 2022-07-27
Dgj 459 change delete submission listener

simensma-fresh:
## Summary
Updated telework form bpmn to latest version.
Changed Delte submission listener to implement the JavaDelegate interface so it can be used as a service task implementation (good practice).

image
image


forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/DeleteSubmissionListener.java
```diff
@@ -18,7 +18,7 @@
 
 
 @Component
-public class DeleteSubmissionListener extends BaseListener implements TaskListener, ExecutionListener {
```
iman-jamali-fw, line 21: You can also delete `TaskListener` and `ExecutionListener` from the imports.

iman-jamali-fw (approved):
Looks good to me!

Issue #430, opened 2022-07-19
Automated backup of Postgres (for Keycloak)


will-freshworks:

Abuchana:
@will-freshworks, Hi Will, can you please confirm this should be a task for the Digital Journeys team as it seems out of context? Thank you!

will-freshworks:
Hi @Abuchana. This is part of the DevOps task list for this project. We needed to get backups of the database for Keycloak automated in case the postgres cluster suffered data loss (very unlikely but better safe than sorry). I actually think @kushal-arora-fw has already finished this task - I will check in with him at our next DevOps stand-up tomorrow morning (PT).

kushal-arora-fw:
That is right, this can be marked done @will-freshworks. Patroni backups are up and running. 

My troubles with Zenhub didn't allow me to change ticket status.

Abuchana:
Thank you! Closed.

PR #489, opened 2022-08-03
Automated backup of Postgres (for Keycloak)

kushal-arora-fw:
## Summary

Issue: Setup automated Backups of Postgres(Patroni) Instance #430 

Enabled automated Postgres backups for Keycloak to support accidental corruption or removal of data to recover from.
* Deployment Config up and running for all environments Dev, Test, and Prod

## Changes

- Added a new GitHub workflow to trigger the build and deployment of the backup Patroni instance
- Base image is a reference from https://github.com/BCDevOps/backup-container as suggested [here - BC DevHub for container backups](https://developer.gov.bc.ca/Backup-Container)
- Removed image trigger for MongoDB backups and this from `automated-db-backups` branch. Once merge to master, dispatch_workflow can be triggered on the appropriate branch
- Changes can be triggered or build using `workflow_dispatch` (once merged) or using github CLI triggers with correct ref

image

## Notes

- If anyone wishes to merge on my behalf - please use `squash & merge` considering there are a lot of adhoc changes pushed to trigger the build and test.


.github/workflows/patroni_backup.yaml
```diff
@@ -0,0 +1,182 @@
...
+      # Build backup container
+      - name: Build backup image
+        working-directory: './deployment/openshift/databases'
+        run: |
+          oc -n ${BUILD_NAMESPACE} process -f prebuilt-db-backup-build.yaml \
+          -p NAME=${NAME} \
+          -p TAG=${BUILD_TAG} | oc -n ${BUILD_NAMESPACE} apply -f - --wait
+
+  # deploy the backup image in Dev
+  deployDev:
```
will-freshworks, line 42: I really don't like this one-job-per-env pattern. Not very DRY. Also you are missing the `if: ${{!github.event.inputs.environment || github.event.inputs.environment == 'dev'}}` guards that the deploy.yml workflow has.

Could you see about merging these into a single `deploy` job and just template in the `github.event.inputs.environment` variable where required to get env-specific behaviour?

deployment/openshift/databases/patroni-backup-deploy.yaml
```diff
@@ -0,0 +1,188 @@
...
+    value: "5432"
+  - name: DB_SERVICE
+    value: patroni-0.patroni
+  - name: DB_SECRET_NAME
+    value: patroni
+  - name: DB_USERNAME_KEY
+    value: app-db-username
+  - name: DB_PASSWORD_KEY
+    value: app-db-password
+  # - name: DB_NAME_KEY
```
will-freshworks, line 159: Remove if not needed

will-freshworks (changes requested):

.github/workflows/patroni_backup.yaml
kushal-arora-fw: Missed this comment before requesting the review again. Kindly hold off until this is resolved.
----

ahh, sure! we can add environment checks - in fact we should 💯 

Also, agree regarding the single`deploy` job, but I was just following the suit there. Will give it a try and get back.

kushal-arora-fw: @will-freshworks - updated PR to make deploy step generic; 

The core of the issue got solved by just naming input to "prod" instead of "production" 🤷‍♂️ ✅

.github/workflows/mongo_backup.yaml
```diff
@@ -50,13 +56,15 @@ jobs:
 
           oc -n ${BUILD_NAMESPACE} start-build bc/${DB_BACKUP_APP} --wait
 
-  # deploy the backup image in Dev
-  deployDev:
-    name: Deploy DB_BACKUP_APP to Dev environment
+  # deploy the backup image
+  deploy:
+    if: ${{contains(['dev', 'test', 'prod'], github.event.inputs.environment)}}
```
will-freshworks, line 61: Isn't this already enforced by the options list?

.github/workflows/patroni_backup.yaml
```diff
@@ -0,0 +1,105 @@
...
+      - name: Build backup image
+        working-directory: './deployment/openshift/databases'
+        run: |
+          oc -n ${BUILD_NAMESPACE} process -f prebuilt-db-backup-build.yaml \
+          -p NAME=${NAME} \
+          -p TAG=${BUILD_TAG} | oc -n ${BUILD_NAMESPACE} apply -f - --wait
+
+  # deploy the backup image
+  deploy:
+    if: ${{contains(['dev', 'test', 'prod'], github.event.inputs.environment)}}
```
will-freshworks, line 53: Enforced by `options`

.github/workflows/patroni_backup.yaml
```diff
@@ -0,0 +1,105 @@
...
+    runs-on: ubuntu-latest
+    env:
+      BUILD_ID: ${{ github.event.number }}
+      BUILD_ENV: ${{ github.event.inputs.environment }}
+      NAMESPACE: d89793-${{ github.event.inputs.environment }}
+      BUILD_NAMESPACE: d89793-tools
+      BRANCH: automated-db-backups
+      NAME: patroni-backup
+      DB_NAME: app
+    needs:
```
will-freshworks, line 64: Probably don't need this `needs` now you only have a single deploy job. You could get rid of the extra checkout action in this case. Also applies to the mongo backup jub

will-freshworks: Oh and the superfluous extra `oc login` command too. To clarify, I mean just merging the steps from `build` into this job

will-freshworks (changes requested):
Good work getting rid of a ton of repetition. Couple of minor comments to help compress things even more.

.github/workflows/mongo_backup.yaml
kushal-arora-fw: It actually does!

.github/workflows/patroni_backup.yaml
kushal-arora-fw: Ohh, `needs` ensure and wait for the build command to be successfully completed. It makes the two jobs sequential. Isn't that the need too?

Also regarding the extra `checkout` and `oc login` you mentioned, my assumption was both these jobs run in their independent containers - so checkout was required. Maybe I am wrong, please let me know if that is not the case and so - these commands are not required?

will-freshworks: > Isn't that the need too?

Yes but since you now only have a single task in `build` and only one job that depends on it, I would just move the build task into the deploy job, rather than have two separate jobs. Doesn't require a new container to be spun up and checkout/login to be run again. Up to you though, I can see an argument for having build and deploy in isolated jobs in the pipeline.

kushal-arora-fw: Got it 👍 

Yeah, but I would like to go ahead with different jobs. A couple of reasons:
1. Separation of concerns
2. Both jobs act on a different namespace. While `build` builds to the tools namespace, `deploy` deploys to the respective namespaces for dev, test, and prod. Thereby keeping them separate would logically make more sense.

Please let me know if you think otherwise.

.github/workflows/mongo_backup.yaml
kushal-arora-fw: Fixed 👍

.github/workflows/patroni_backup.yaml
kushal-arora-fw: Fixed 👍

will-freshworks (approved):

Issue #481, opened 2022-07-29
TELEWORK: Workflow becomes detached from form 


DonMoffatt:
Background:
- [ ] On Thursday July 28 the @DonMoffatt and others on the PSA discovered that the workflow is becoming detached if a forms designer edits the form in PROD. This happens even if we are not attaching the workflow. We alerted Simen, who quickly reproduced the problem and found a temporary workaround.
 
Steps to Reproduce:
- [ ] Put the form in edit mode. Exit without making changes (or makes changes - it happens both ways). 
 
Actual Behavior:
- [ ] The workflow becomes detached upon saving. After saving the workflow is not called, and when you put the form in edit mode the workflow is no longer attached.
 
Expected Behaviour:
 - [ ] The workflow does not become detached when you save the form after an edit by a forms designer. 
Environment:
- [ ]  PROD

MeghanStothers:
Only happens in Prod

MeghanStothers:
@simen can we/dev team connect about a longer term fix for this one?

iman-jamali-fw:
@MeghanStothers the fix for this issue is in PR review.

iman-jamali-fw:
Deployed and tested successfully on DEV.

iman-jamali-fw:
Deployed and tested successfully on TEST. Will move the ticket to Review to be tested by the PSA team.
The telework form was cleaned in the TEST's DB.

iman-jamali-fw:
@DonMoffatt @Stella-Archer 
This is ready to be reviewed on TEST. If it was successful please let me know to move it to PROD.

DonMoffatt:
Hi @iman-jamali-fw  Hasan and I tested this in TEST. Please apply the fix to PROD. Can  you also tell us what broke after it appeared to work on PROD before 10? Please IM me/Hasan or Stella if something critical like that breaks in addition to tagging us here.

MeghanStothers:
Updated to high priority

iman-jamali-fw:
@DonMoffatt @HasanSifat It is deployed to PROD. Can you test to see if the issue resolve?

Cleanup on PROD requires more work, I'll work on a strategy and will discuss it with @simensma-fresh before moving forward. In the meantime, the issue of workflow detached on form update/save should never happen again with this fix.

[...]

PR #493, opened 2022-08-05
form-process-mapper returns only the first record in db to fix the is…

iman-jamali-fw:
## Summary
This PR fixes the issue of `form-process-mapper` leakage. As reported in #481

## Background:
There was an issue that in some cases during the editing a form by a designer, a new form-process-mapper was created in the DB instead of updating the one designer changed.

Next time the designer wants to edit a form, since there are multiple entries of `form-process-mapper` in DB, another one, different from the one last time was changed, was loaded, giving the designer the impression that previous changes were not saved properly, which in fact it was saved to another `form-process-mapper`.

## FIX
This PR guarantees that only the first item in the DB is returned every time a `form-process-mapper` is fetched by `form-id`. This means the designer will always work with the same entry and changes will be properly saved.

## Note

- The actual issue of leakage was fixed in [this PR](https://github.com/bcgov/digital-journeys/pull/486)
- In the next version of the form-flow-ai, form-process-mapper will be saved as a separate entry with each change with a new version. The latest version of that mapper will be loaded. Before that this pr will fix our issues.

### DIFF
--- a/forms-flow-web/src/components/Form/Item/Edit.js
+++ b/forms-flow-web/src/components/Form/Item/Edit.js
@@ -200,10 +200,6 @@ const Edit = React.memo((props) => {
                 data["processName"]= processListData.processName;
               }
               
-              // keep the existing status
-              if (processListData.status) {
-                data["status"]= processListData.status;
-              }
 
               const updated =
                 processListData && processListData.id ? true : false;
--- a/forms-flow-api/src/formsflow_api/models/form_process_mapper.py
+++ b/forms-flow-api/src/formsflow_api/models/form_process_mapper.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 from http import HTTPStatus
 
-from sqlalchemy import UniqueConstraint, and_, desc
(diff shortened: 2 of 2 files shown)


forms-flow-web/src/components/Form/Item/Edit.js
```diff
@@ -200,10 +200,6 @@ const Edit = React.memo((props) => {
                 data["processName"]= processListData.processName;
               }
               
-              // keep the existing status
-              if (processListData.status) {
-                data["status"]= processListData.status;
-              }
```
iman-jamali-fw, line 206: Turned out this was not necessary. Removing to keep the code cleaner.

david-sabadash (approved):

simensma-fresh (approved):

forms-flow-web/src/components/Form/Item/Edit.js
HasanSifat: Noted! thanks for fixing this bug Iman :)

PR #506, opened 2022-08-15
Added daily backup trigger for forms-flow-files PVC

kushal-arora-fw:
## Summary

- Added github workflow cron trigger to backup form-flow-files PVC
  - Schedule - Daily at 9am
  - Source: forms-flow-files pvc
  - destination: forms-flow-files-backup

### DIFF
--- a/.github/workflows/form-flow-files-backup.yaml
+++ b/.github/workflows/form-flow-files-backup.yaml
@@ -1,0 +1,40 @@
+name: Run file backups for forms-flow-files
+
+on:
+  schedule:
+    - cron: '0 9 * * *'
+
+jobs:
+  fileBackup:
+    name: Trigger file backups
+    strategy:
+      fail-fast: false
+      #Run file backups for multiple environments
+      matrix:
+        namespace: [d89793-dev, d89793-test, d89793-prod]
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


kkapoor-fresh (changes requested):
-I believe the form-flow-files is created as a stateful set in one of the namespaces whereas it is deployed as a deployment config in others. We just need to make sure that the "oc exec" command works fine while looping over the namespace matrix.  Could you please confirm if the same code will work for all namespaces?
- Also, we can reduce the no. of file back backups later to 2-3 once we have all the data from the original pvc into the new pvc. The file back up code will copy the entire data to the new pvc on a daily basis, so we will anyway have the required data in just a day of backup export.

kushal-arora-fw:
> -I believe the form-flow-files is created as a stateful set in one of the namespaces whereas it is deployed as a deployment config in others. We just need to make sure that the "oc exec" command works fine while looping over the namespace matrix. Could you please confirm if the same code will work for all namespaces?
> 
> * Also, we can reduce the no. of file back backups later to 2-3 once we have all the data from the original pvc into the new pvc. The file back up code will copy the entire data to the new pvc on a daily basis, so we will anyway have the required data in just a day of backup export.

- addressed ✅

kkapoor-fresh (approved):
Approved

PR #516, opened 2022-08-16
updated formio-upload docker - removed grsync

kushal-arora-fw:
## Summary

updated formio-upload docker - removed grsync

Build fix


kushal-arora-fw:
DIscussed on slack. Merging.

PR #524, opened 2022-08-19
fix redundant backup for forms-files-flow when multiple pods

kushal-arora-fw:
## Summary

Problem: There were redundant backups getting created when multiple pods running.

image

Fix: removed unnecessary loop


kkapoor-fresh (approved):
Approved

PR #526, opened 2022-08-22
added mongo deploy workflow

kushal-arora-fw:
## Summary

Added HA Mongo CI/CD workflow

to avoid synchronization the future issues with changes done directly in the console vs via code.

Intent -> No changes be done directly in console in the future

### DIFF
--- a/.github/workflows/mongo-deploy.yaml
+++ b/.github/workflows/mongo-deploy.yaml
@@ -6,7 +6,6 @@ on:
       environment:
         description: 'Environment'
         required: true
-        default: 'dev'
         type: choice
         options:
         - dev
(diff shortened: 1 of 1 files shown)


kkapoor-fresh (commented):
I had a discussion with Simen and he recommended to have all these services triggered by just a workflow dispatch even (manually). he mentioned while the level of automation that they need is  not necessarily a full fledge automated CI/CD pipeline which gets trigger as soon as A PR is merged  but rather a "clickable" button available in the Actions console via a workflow dispacth event so that they can deploy these services whenever they need to.

kushal-arora-fw:
> I had a discussion with Simen and he recommended to have all these services triggered by just a workflow dispatch even (manually). he mentioned while the level of automation that they need is not necessarily a full fledge automated CI/CD pipeline which gets trigger as soon as A PR is merged but rather a "clickable" button available in the Actions console via a workflow dispacth event so that they can deploy these services whenever they need to.

Removed default workflow dispatch to "only" trigger manually

kkapoor-fresh (approved):

PR #527, opened 2022-08-22
Print Confirmation Popup

david-sabadash:
## Summary
added success toast to Print As PDF button for forms

## Changes
Updated the onClick for the Print As PDF button to also trigger a toast

image


iman-jamali-fw (approved):
LGTM

Issue #468, opened 2022-07-27
Validation Error is not visible to manager submitting


Stella-Archer:
Background: As a manager I am trying to submit my decision on a telework agreement. I don't notice or see the 'agree' button. 
Nothing appears to happen when I submit as the validation error is happening in an area I can't see on the screen
 
Steps to Reproduce: process a telework form for more than 3 days, manager does NOT select 'agree' but does everything else correctly, then submits... it look like it's broken and nothing happened
 
Actual Behavior: Manager does not see the validation error on the area of the screen where they are working
 
Expected Behaviour: Manager should see the validation error on the area of the screen where they are working
 
Environment: ALL![Validation Message is not on submit screen.gif](https://images.zenhubusercontent.com/61e5ab4a9a5a1e4bd27a92b8/ca6e335b-ee39-48ce-a0a0-7fea4c652463)

Stella-Archer:
![image.png](https://images.zenhubusercontent.com/61e5ab4a9a5a1e4bd27a92b8/de53b6b8-fb17-4203-bac0-d653b8ed4d78)

simensma-fresh:
Would a solution where we scroll you to where the error message is be OK here? 

@david-sabadash If so, There might be some built in functionality we can use for this: https://github.com/formio/formio.js/issues/3650

david-sabadash:
Investigation ccomplete on this task; currently the component we need access to in order to implement this functionality is not exposed to us in our implementation. None of the currently available properties of the Form component accurately allow us to catch validation events and focus them on the screen. In order to implement this functionality, a full reimplementation of the Form component would be needed. This is related to @simensma-fresh 's suggestion - the class method mentioned in the formio repo is only accessible in the core JS library, and not in the react-formio library we currently have implemented. we would need to switch over to the formio.js library, or alternatively, make a local override for the Form component similar to how we are handling the existing forms-flow-ai component overrides.

MeghanStothers:
Decision to keep it open at today sprint review - Simen might have a workaround and will connect with David.

Another option is a pop-up message
Another option is the red box at top of page that lists errors.

HasanSifat:
Tested successfully in DEV. Ready to be transferred to TEST

Stella-Archer:
[...]

PR #532, opened 2022-08-24
scroll to validation error field was added to form edit, form edit pa…

iman-jamali-fw:
## Summary
This PR fixes the issue with the validation error that was not scrolled to during the supervisor/manager signoff. The problem was described in #468 

## Changes
- CSS rules were added to the global style to fix the issue with two scrollbars on form edit submission page (for example in the Telework form, the page where the supervisor signs off the form)
- Reference to formio instance was added in Edit submission page to listen and find the validation errors
- Logic to scroll to the top most form field with validation error was added

## Notes
Please note that the formio instance has only become available after a certain amount of time. As a workaround, I added an interval to check for the formio instance every 1000 ms with cleaning logic. Please let me know if there is a better/cleaner way to handle this.

### DIFF
--- a/forms-flow-web/src/components/Form/Item/Submission/Item/Edit.js
+++ b/forms-flow-web/src/components/Form/Item/Submission/Item/Edit.js
@@ -52,9 +52,9 @@ const Edit = React.memo((props) => {
 
   let scrollToErrorInterval = null;
   useEffect(() => {
-      scrollToErrorInterval = setInterval(() => {
-        scrollToErrorOnValidation()
-      }, 1000);
+    scrollToErrorInterval = setInterval(() => {
+      scrollToErrorOnValidation()
+    }, 1000);
     return () => {
       clearInterval(scrollToErrorInterval);
     }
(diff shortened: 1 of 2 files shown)


forms-flow-web/src/components/Form/Item/Submission/Item/Edit.js
```diff
@@ -49,6 +50,16 @@ const Edit = React.memo((props) => {
     }
   }, [applicationStatus, userRoles, dispatch, submissionId, formId, onFormSubmit ]);
 
+  let scrollToErrorInterval = null;
+  useEffect(() => {
+      scrollToErrorInterval = setInterval(() => {
```
simensma-fresh, line 55: I think we might be able to use either `useEffect` (or potentially `useCallback` since we're dealing with a ref) here to trigger the listener setup instead of an interval

iman-jamali-fw: Well, I tried different combinations of `useEffect` and `useCallback`. The problem is that even if after listening to `refForm`, it becomes available (i.e. not undefined), still the formio instance inside `refForm.current` if undefined at least after 1s. If my code touches that before, the form won't be responsive anymore and the user has to refresh the page.

Apart from that, I see some weird behavior in the `useEffect` in this project, not sure what's different from other React apps that adding a dependency to `useEffect` or `useCallback` won't work as expected, other words, the call back doesn't trigger when a dependency changes, like in other React apps.

simensma-fresh (approved):

PR #544, opened 2022-08-26
Forms flow files CI/CD

kushal-arora-fw:
## Summary

This activity is done as part of synchronization efforts between the code and the console; and to avoid any future irregularities.

Changes in the PR include:
- Added forms-flow-files Build Config
- Added forms-flow-files Deployment Config
- Added forms-flow-files GA Workflow on dispatch

Note: Updated all config with the values from the prod namespace.

Also, forms-flow-files in dev are currently running via a Deployment Config, while Test and Prod environments run as StatefulSets. As part of this change, all envs will run as StatefulSet for consistency.

----
~~Pending: There are some certificates added to the application. Figuring out a way to include them. Here's the screenshot of the pending resources:~~
image
Confirmed with Simen -> above resources are managed by Openshift. So no action needed


kkapoor-fresh (approved):
Looks Good to me! Just ensure that the parmeter values passed in do not inadvertently change the current config of the app running in its current state.

deployment/openshift/form_flow_files_dc.yaml
```diff
@@ -0,0 +1,232 @@
...
+objects:
+  - kind: PersistentVolumeClaim
+    apiVersion: v1
+    metadata:
+      name: ${NAME}-data
+      labels:
+        app: ${NAME}
+    spec:
+      accessModes:
+        - ReadWriteMany               # this is ReadWriteOnce currently
```
simensma-fresh, line 100: I'm wondering what will happen if we change this from `ReadWriteOnce` to `ReadWriteMany` as I believe it is immutable?
Will it replace the old volume and all the data within it? As we have to make sure that we retain the data that's already there

kushal-arora-fw: Great catch! I had the same question and thought of checking this with you, but it got skipped completely. I added it for deployment failures I was facing while modifying pod definitions.

> Will it delete the underlying persistent volume as well? I am not very sure.

With that in my mind, let me change it back to RWO (as in console currently) and change it later, if needed.

simensma-fresh: Sounds good! Yeah we would want to change it to RWM fairly soon so we're not limited to just one pod, but sounds good to get this merged first

kushal-arora-fw:
@simensma-fresh - There's one more difference in this PR to note, which is different from the console. Currently `forms-flow-files` is running as a `DeploymentConfig` in dev whereas it is a `StatefulSet` in test and prod.

For consistency in this PR and to avoid unnecessary complexity, I have made it to StatefulSet. So - if we deploy this, it will run as StatefulSet in dev as well. Hope that's ok. [Also, once deployed - do we first need to remove the existing DeploymentConfig in dev?]

simensma-fresh (approved):

simensma-fresh:
@kushal-arora-fw Changing it to a StatefulSet makes sense to keep it consistent with the other environments. I would guess that we would indeed have to delete the Deployment config in dev

fazil-ey:
@simensma-fresh is this still a relevant ticket? Can we close this?

Issue #491, opened 2022-08-04
TOOL: External links


MeghanStothers:
**User Story**
“As a Forms Designer, I want all links to open in a new window, so I don't need to hand code this in forms flow, and I can be sure users are not popped out of the form when they check out a link.” 

**Acceptance Criteria**
All links open in a new window
in all environments

Stella-Archer:
Anchor links?

iman-jamali-fw:
@Stella-Archer The changes are on DEV and TEST for your testing.

Testing instruction:
- Please open this form on [DEV](https://digital-journeys-dev.apps.silver.devops.gov.bc.ca/form/630e533b1c6951caa7ca6559) and [TEST](https://digital-journeys-test.apps.silver.devops.gov.bc.ca/form/630e65dcdb0b297ffa5bfd1d)
- You'll see two links, first configured to be opened in a new tab, and second, configured to be opened on the same page.
- Clicking on any of the two links will open a new tab.


Please note, that this fix converts all the links on the forms to be opened in a new tab. Per the need for an internal link, we can upgrade this solution to support internal links (for scrolling/navigation within the same form).

HasanSifat:
Tested successfully in TEST and DEV. Ready to be transferred to PROD

iman-jamali-fw:
Another fix was implemented to only affect the forms links and not the TOOL. Also, the new fix, makes it easier to support internal links for navigation if needed in the future.

MeghanStothers:
Iman to push to test and assign to Hasan to test

iman-jamali-fw:
@HasanSifat The new fix is deployed to DEV and ready for testing. If you could, please test the links in the Telesowrk form when filling the form and when the manager sign-off as the new fix makes changes to the form. On top of that, you can test it [here](https://digital-journeys-dev.apps.silver.devops.gov.bc.ca/form/630e533b1c6951caa7ca6559).

Thank you

HasanSifat:
Tested successfully in DEV. Ready to be transferred to TEST.

iman-jamali-fw:
@HasanSifat Deployed to TEST and ready for the test.

HasanSifat:
Tested successfully in TEST. Ready to be transferred to PROD

PR #547, opened 2022-08-30
base html element was added to open all the link in a new tab

iman-jamali-fw:
## Summary

This PR will force all the `<href>` tags to open in a new tab as requested in #491 

## Changes
- A `<base>` tag was added to the `header` of the topmost index overwriting all the `<href>` tags to open in new tabs.


simensma-fresh (approved):
🔗

PR #548, opened 2022-09-01
making the form links open in a new tabs

iman-jamali-fw:
## Summary

This PR removes the global `<base>` tag that makes all the `<href>` tags in the app open in new tabs and instead, add that attribute (`target="_blank"`) only to form components with an `<href>` tag. #491 

## Changes
- Remove the global `base` tag
- Add a form util function that read through the form components and make the links open in new tabs
- Refactored the form util functions that rely on formio object in a separate module.


bhumin-fw (approved):
Looks good to me.

simensma-fresh (approved):

MissionT:
A new Issue with links not opening to a new page again, possibly from the 5.0 updated. Devs have a global function for all forms to open to a new page. Incident was noticed when building the 1.10 Complaint form.

Issue #568, opened 2022-09-09
Date field not carrying over from employees submission to supervisors copy =>  3. Date/duration of agreement.


MissionT:
Background:
We have received inquiries again regarding the Date field appearing empty when the supervisor receives the employees copy.
The date is visible in the employees version but not when it reaches the supervisor/manager. 
 
Steps to Reproduce:
Tested as a supervisor receiving both a 2 day and 3 day submission for approval.  

 
Actual Behavior:
 Both emails request came in a timely matter and were approved with the date field data appearing.

Expected Behaviour:
Appears to be hit and miss.  We have two emails in the inbox. One at 10:25 am from Thursday, and another at 4:18 pm, both are missing the Date field that the employee  filled out. Date field is filled out correctly and sent (since it is a mandatory field).  The issue is the date does not always carry over to the supervisors copy to approve (null field). 

 
Environment: Prod 
[Uploading... JenniferDouglasDateMissing.msg]
[Uploading... DanieleWilliamsonDateMissing.msg]

MeghanStothers:
Still under investigation - we are getting reports of some user cases where this is happening. 
Date implementation is still there in service side pdf generation.

iman-jamali-fw:
The issue was that the `minDate` validation rule was triggered during the pdf generation and didn't let the date value passes to those fields. It only happened when a submission was signed off one or more days after the original employee submission.

It was fixed by in the pdf generation server by removing the `minDate` validation rule.

iman-jamali-fw:
@HasanSifat Deployed to DEV, ready to be tested.

Abuchana:
Simen to push to Prod

simensma-fresh:
@HasanSifat This has now been deployed to production

HasanSifat:
tested successfully in PROD! Ready to be closed

PR #583, opened 2022-09-16
date fields minDate validation were removed before pdf server generation

iman-jamali-fw:
## Summary (ticket #568)

This PR will fix the issue with the date field of the generated PDF files that were rendered empty when a supervisor signed off a submission one or more days after the original employee submission.

## Changes
- A function was added to remove the `mindDate` of date fields before the form object is sent to the pdf is generated.
- Formatted the generate file to look nicer :)

## Notes
I tried different approaches to disable the date validation such as making those fields `readOnly`, similar to what I did on the web form, and also various ways to use formio built-in functions --in the form generation HTML, but none worked. Finally, changed the form object before it passed to the PDF generation script in Puppeteer.

### DIFF
--- a/formio-upload/pdf-generation/formUtils.js
+++ b/formio-upload/pdf-generation/formUtils.js
@@ -1,0 +1,28 @@
+/*
+* A recursively lookup in all form components and set the target value
+* for the target property with the specified key
+*/
+function setValueForEveryComponentByKey({
+  components,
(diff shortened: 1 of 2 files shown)


bhumin-fw (approved):
Looks good to me

simensma-fresh (approved):

PR #622, opened 2022-10-05
DGJ 598 draft page remove draftId and 607 remove application status filter

bhumin-fw:
## Summary
DGJ 598 change column value(Draft name, showcase employee name along with form name) on the draft page
DGJ 607 remove the filter from the application status column on the application page

### DIFF
--- a/forms-flow-web/src/containers/Head.js
+++ b/forms-flow-web/src/containers/Head.js
@@ -9,9 +9,7 @@ const Head = React.memo((props) => {
     return (
       <div className="header-container">
         <div className="main-header">
-          <div
-            className={`head-item`}
-          >
+          <div className="head-item" style={{ color: "black" }}>
             <h3 className="application-head application-head-active" style={{marginBlockEnd: "5px"}}>
               <i
                 className={`fa fa-${item?.icon}`}


forms-flow-web/src/components/Draft/table.js
```diff
@@ -57,8 +59,14 @@ function timeFormatter(cell) {
   return <label title={cell}>{localdate}</label>;
 }
 
-const nameFormatter = (cell) => {
-  const name = startCase(cell);
+const nameFormatter = (cell, row) => {
+  let employee = '';
+  if (cell.toLowerCase().includes('sl review')) {
```
iman-jamali-fw, line 64: Sooner or later, we need to define env and constant for the form names and rely on that instead of hardcoding the names.

bhumin-fw: That make sense, That's why I did use .includes() for now. Later we will update it.

bhumin-fw:
@iman-jamali-fw Good catch, I have reverted removed the code, and added a comment with the issue link.

simensma-fresh (approved):

Issue #570, opened 2022-09-12
SL Review - Post Submission Data hand-off to ODS


fazil-ey:
**User Story**
As an ADM, once I have filled in an application I want the data to be transmitted to ODS so that the SHRs have the most updated report to present to execs

**Tasks**
- [x] confirmation message to ADMs after clicking on the button (this is now addressed in ticket#621) 
- [x] Data on form submission gets transmitted to ODS

**Acceptance Criteria**
When the ADM is able to data in the SL form before the validation meeting and also able to come back make changes and re-submit again

iman-jamali-fw:
Hi Patrick @T854236,
Can you help me with this task of sending SL Review data to ODS?

Please let me know what you need from me to have this data stored in an ODS endpoint?

I'd imagine that you need a table of the fields (name, type, require/optional) from me? Maybe in an excel sheet like the one for Telework.

And do you think, similar to Telework, two endpoints for test and prod are good enough?

T854236:
@iman-jamali-fw 
Hi Iman,

yes please provide list of columns names, data types, length for each, allow nulls or not.  Excel sheet will work great.  Yes I think an endpoint to start for each environment should work.

HasanSifat:
Hi @iman-jamali-fw and @T854236 
The requirement to send data to ODS will remain unchanged. However, there might be a change in the requirement for a hard button. We plan on discussing this with Simen tomorrow morning about the current process map and how it may be implemented in the tool before you put in too much work on this. The fields and their responses are somewhat finalized and so you can go ahead make the spreadsheet. But as I said, the requirement to send data off to ODS for reporting still remains.

iman-jamali-fw:
@T854236  Here is the initial list of SL Review fields. I wasn't sure where to create a spreadsheet and how to show nested objects, so adding the values here. Feel free to update this as the form evolves. I'm not sure about length and allowing for nulls, maybe @simensma-fresh  can comment on that.

@HasanSifat The SL Review form v2 on DEV fields (API names) were updated with these names. I updated some of your logic to not break the validation with the new names. If you can , please check the form to see if everything works as before. Thanks

```json
{
  "applicationId": "string",
  "applicationStatus": "string",

  "employeeName": "string",
  "employeeNumber": "string",
  "employeePositionTitle": "string",
  "employeeWorkCity": "string",
  "ministry": "string",
  "division": "string",
  "branch": "string",
  "admSupervisorName": "string",
  "admSupervisorEmail": "string",

[...]

PR #623, opened 2022-10-05
DGJ-570 SL Review workflow updated, Triggers Camunda on application edit, flatten data on send to ODS

iman-jamali-fw:
## Summary

This PR (Ticker #570) enables the SL Review resubmission through the `application` tab and sends its data to ODS on each update.

## Changes
- SL Review workflow upated
- Logic to trigger camunda workflow under application edit was added
- The confirmation page for SL Review submit/resubmit was added

## Notes
Please note that the sendToODS listener in Camunda still needs some work including flatMap form data.


forms-flow-web/src/components/Form/Item/Submission/Item/Edit.js
```diff
@@ -146,6 +177,34 @@ const Edit = React.memo((props) => {
     return <Loading />;
   }
 
+  const onApplicationFormSubmit = (actionType = "") => {
```
simensma-fresh, line 180: I'm wondering if we can change this to be decoupled from the form "submit" button and instead be triggered through a Form Event to match how this happens on the Task page (ServiceTaskDetails.js -> onCustomEventCallBack). That way we can easily disable / enable this per form as needed and also have the option to pass along custom data to the Camunda Task you want to progress. You can check the "Submit Decision" button on the Telework form for that logic

```
  form.emit('customEvent', {
            type: "actionComplete",   
            component: component,
            actionType:data.action
      }); 
```

simensma-fresh (changes requested):
Overall looks good! Two suggestions on how to make this a bit more reusable and flexible

forms-flow-web/src/components/Form/Item/Submission/Item/Edit.js
simensma-fresh: And if you're able to add the ticket number to the title of the PR, that would be great! Just makes it easier to reference this if needed in the future

forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/SendSubmissionToODSDelegate.java
```diff
@@ -53,22 +55,23 @@ private void sendSubmissionToODS(DelegateExecution execution) throws IOException
         Object managerIdir = execution.getVariable("manager_idir");
         Object managerGuid = execution.getVariable("manager_guid");
 
+        // Todo: Replace the form name with an env
+        boolean hasNestedObjects = formName.equals("SLReview");
```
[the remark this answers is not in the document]
iman-jamali-fw: addresed

forms-flow-web/src/components/Form/Item/Submission/Item/Edit.js
iman-jamali-fw: addressed.

PR #634, opened 2022-10-12
DGJ 604 remove application id column change application name

bhumin-fw:
## Summary
Major Changes
- Remove the ApplicationId column
- Add the employee name along with the form in the form-name column under the application tab.
- Change the word "Application" to" Form" under the application tab.

Minor fixes
- Change the description of the Application, Draft, and Form tabs.
- Send blank array in employee search instead of 404, so the Select component keeps triggering search API.

## Changes
- Add fetch submission code in form-flow-API in the GET applications API.

## Notes
Move the select EmployeeName code to the helper file from the application/draft table.js file. Use keywords like "SL review" and "telework" to identify the form name.
Later when we finalize constants for each form, we will change this small pic of the code with the exact form name check.


forms-flow-api/src/formsflow_api/services/application.py
```diff
@@ -506,3 +507,21 @@ def get_application_count(auth, token, **kwargs):
...
+        applications = []
+        try:
+            formio_service = FormioService()
+            form_io_token = formio_service.get_formio_access_token()
+            for a in application_list:
+                submission = formio_service.get_submission( {"form_id": a["formId"], "sub_id": a["submissionId"]}, form_io_token)
+                if submission:
+                    a["submission"] = submission
+                    applications.append(a)
+        except BusinessException as err:
```
iman-jamali-fw, line 523: I would `raise` the exception in the service and handle it in the application resource.

forms-flow-web/src/components/Application/table.js
```diff
@@ -101,29 +105,31 @@ export const columns = (
...
-        placeholder: `\uf002 ${t("Application Id")}`, // custom the input placeholder
-        caseSensitive: false, // default is false, and true will only work when comparator is LIKE
-        className: "icon-search",
-        style: customStyle,
-        getFilter: (filter) => {
-          idFilter = filter;
-        },
-      }),
-    },
+    /*commented below code, for more detail visit below link
```
iman-jamali-fw, line 108: I really like your approach of leaving the pr link for the commented line. I'll follow that in my future prs :)

iman-jamali-fw (approved):
LGTM!
Left one comment.

forms-flow-api/src/formsflow_api/services/application.py
bhumin-fw: @iman-jamali-fw a thought behind not raising an exception here is. It is s sub-information that we are fetching. The issue with the connection of formio service will break the loading application page.
If we do not raise an exception it will at least load all the applications and the user will not impact here.

I will love to raise an exception if that's the case that seems reasonable. or If you find any point over my thought process please feel free to let me know here,

iman-jamali-fw: OK, that makes sense.

PR #647, opened 2022-10-19
DGJ-572 Fix employee search by name issue with special characters

bhumin-fw:
## Summary

Fix special character bug in search employee name ODS API.

## Changes
Replace the special character with _, so it will work as an SQL-like query and we can overcome with ODataPaser error from the ODS APIs.

### DIFF
--- a/forms-flow-api/src/formsflow_api/services/employeeDataService.py
+++ b/forms-flow-api/src/formsflow_api/services/employeeDataService.py
@@ -1,4 +1,6 @@
 import requests
+import re
+from urllib.parse import unquote
 from http import HTTPStatus
 from flask import current_app
 from formsflow_api_utils.exceptions import BusinessException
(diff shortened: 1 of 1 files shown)


forms-flow-api/src/formsflow_api/services/employeeDataService.py
```diff
@@ -54,9 +58,8 @@ def get_employee_names(args):
                        headers={"Authorization": current_app.config.get("ODS_AUTH_TOKEN")})
 
       except Exception as e:
-        raise BusinessException(
-          {"message": "Failed to get employee names from ODS"}, HTTPStatus.INTERNAL_SERVER_ERROR
-        )
+        current_app.logger.error(e)
```
bhumin-fw, line 61: I have omitted the error raise statement.
any error is stop search functionality in the Form "remote select" component. 

I have logged an error here, So if we receive any query from QA or User we can take look and see what new case we have to handle here to search from the ODS.

iman-jamali-fw (approved):
Looks good to me

simensma-fresh (approved):

PR #661, opened 2022-10-24
DGJ-636 Hide senior leadership review form the task list tab

bhumin-fw:
## Summary
DGJ-636 Hide the Senior Leadership Review form from the task tab

## Changes
Added constant that helps to filter tasks. It must be added as a description on the Camunda process "User Tasks".

## Notes
N/A


iman-jamali-fw (approved):
LGTM!

PR #688, opened 2022-11-02
DGJ-673 Release note popup for the PROD release

bhumin-fw:
## Summary
DGJ-673 Added a new model for the release note. It will map users and release notes. And only show unread release notes on the /form page.
DGJ-676 Enable the back button on the form page.

## Changes
- Added APIs for admin/designer role to add new release notes for all users. (It will user by DEV for now)
- Added API to fetch unread release notes for the user and mark it read when the user closes the release-note modal popup

Release-note

## Notes
- Omit all the errors for retrieving release-note and save read release notes. This service will run in the background and we do not need to show any toast message to the user regarding its process/success/error.


forms-flow-web/src/components/Form/List.js
```diff
@@ -561,6 +580,30 @@ const List = React.memo((props) => {
...
+        </Modal.Body>
+        <Modal.Footer>
+          <Button type="button" className="btn btn-default" onClick={onClose}>
+            <Translation>{(t) => t("Close")}</Translation>
+          </Button>
+        </Modal.Footer>
+      </Modal>
+  );
+};
+
```
iman-jamali-fw, line 606: I would move this component to its own file. Maybe in `components/Form`?

forms-flow-web/src/components/Form/List.js
```diff
@@ -345,6 +355,15 @@ const List = React.memo((props) => {
 
   return (
     <>
+      <ReleaseNoteModal 
+        showModal={showReleaseNoteModal} 
+        onClose={() => {
+          setShowReleaseNoteModal(false);
+          dispatch(readReleaseNote({release_note_id: releaseNoteData.data?.id}));
+          }}
```
iman-jamali-fw, line 363: Just make sure after it is read (`onClose`), the unreadNote is removed from the state (by resetting or fetching unread again), so if a user returned to this page, the modal does not pop up.

iman-jamali-fw (commented):
Looks great! Huge work 👏🏼
Only had two comments.

forms-flow-web/src/components/Form/List.js
```diff
@@ -561,6 +580,30 @@ const List = React.memo((props) => {
...
+        onHide={onClose}
+        size="lg"
+        >
+        <Modal.Header style={{color: "#fff", backgroundColor: "#036"}}>
+          <Modal.Title>
+            <h3>{title}</h3>
+          </Modal.Title>
+        </Modal.Header>
+        <Modal.Body>
+            <div dangerouslySetInnerHTML={{ __html: content }} />
```
simensma-fresh, line 596: Thoughts on rendering this using the `<Form` component from `react-formio`? As I think that internally does some sanitization etc.
That way we would also be one step closer to moving this into the platform for the PSA team to manage as in the future let them use the form builder to create the notices.

forms-flow-web/src/components/Form/List.js
bhumin-fw: Moved it, Keep useEffect in the List.js as with the new suggestion to render form(using formio supported component) break when form unloads when hiding modal.

forms-flow-web/src/components/Form/List.js
bhumin-fw: Correct Iman, The ReleaseNote object is set to blank when clicking on "onCLose". even if API requests succeed or fail.

forms-flow-web/src/components/Form/List.js
bhumin-fw: @simensma-fresh I have updated API and frontend to support `<Form ` from the `react-formio` . Now in place of HTML string, we will store FORM JSON (generated from the form builder, please check MS team for new database row) in the database, and render it on the release note Modal using `<Form`.

simensma-fresh (approved):
Looks great!

iman-jamali-fw (approved):

PR #711, opened 2022-11-10
feature/DGJ-680_floating-button

iman-jamali-fw:
## Summary

This PR would allow a Form designer to add floating buttons to the forms.

## Changes
- Adding floating styles that a designer can use to make a button float on two positions (more can be added per request).
- Make the PDF button a custom form.io button that emits the print event to the different components.
- Adding the necessary logic to hide all floating buttons during the PDF generation on FE.
- Make the `PrintPDF` button enable and `saveAsDraft` hidden on `submission.view` page (only applicable to Telework) 

## Notes

The floating button will be positioned on top of the page on the server-generated PDFs. I didn't spend time removing them from that pdf since we might replace it with the AOT PDF generator in the future. See below image:
Screen Shot 2022-11-09 at 5 53 01 PM

Screen Shot 2022-11-09 at 5 48 50 PM
Screen Shot 2022-11-09 at 5 49 06 PM
Screen Shot 2022-11-09 at 5 49 28 PM
Screen Shot 2022-11-09 at 5 49 42 PM

### DIFF
--- a/forms-flow-web/src/helper/formUtils.js
+++ b/forms-flow-web/src/helper/formUtils.js
@@ -27,4 +27,21 @@ const scrollToErrorOnValidation = (formio, scrollToErrorInterval) => {
   }
 };
 
-export { convertFormLinksToOpenInNewTabs, scrollToErrorOnValidation };
+const enableFormButton = (formio, enableButtonInterval, buttonKey, onClick) => {
+  if (formio) {
+    clearInterval(enableButtonInterval);
+    const button = document.getElementsByName(`data[${buttonKey}]`)[0];
+    if (button) {
+      button.removeAttribute("disabled");
+      if (onClick) {
+        button.onclick = onClick;
+      }
+    }
+  }
+};
+
+export {
+  convertFormLinksToOpenInNewTabs,
+  scrollToErrorOnValidation,
+  enableFormButton,
+};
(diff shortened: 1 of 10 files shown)


forms-flow-web/src/helper/formUtils.js
```diff
@@ -27,4 +27,21 @@ const scrollToErrorOnValidation = (formio, scrollToErrorInterval) => {
   }
 };
 
-export { convertFormLinksToOpenInNewTabs, scrollToErrorOnValidation };
+const enableFormButton = (formio, enableButtonInterval, buttonKey, onClick) => {
```
iman-jamali-fw, line 30: the usual emit event doesn't work on the `submission.view` since the form is `readOnly`. As a workaround wrote this helper.

bhumin-fw (approved):
Looks good to me. Awesome work here Iman, It is way more simplified than I thought :).

simensma-fresh (approved):
Looks good! Just one suggestion, I think we can fairly easily also hide it from the server side generated PDFs by adding a css tag to hide `. floatingButton`s in the template used for generation (formio-upload/pdf-gteneration/formio.html)

iman-jamali-fw:
> Looks good! Just one suggestion, I think we can fairly easily also hide it from the server side generated PDFs by adding a css tag to hide `. floatingButton`s in the template used for generation (formio-upload/pdf-gteneration/formio.html)

Good one, will try that before merging.

PR #737, opened 2022-11-22
DGJ-723 web and api as separate deployment files

bhumin-fw:
## Summary

DGJ-723 Web APP and  Web API as separate deployment files

## Changes
Added 2 separate deployment files, one for the web and one for API.

## Notes
in the image, I have mentioned formflow/forms-flow-web, I'm not sure here, do we need to add the prefix formsflow/ here or not. Let's review it and fix that image value.


kkapoor-fresh (commented):
Just curious over a few things here:
1. Could you please share the specific need to separate the web and api deployments? Are we looking for keeping the deployments independent of each other? 
2. We are using the "matrix" property but the workflow doesn't appear to have run multiple times with a list of options. Is it just to parameterize the workflow where you just would have the change the image details in the matrix section if you want to run it with a different image
3. I am keen to know the reason to replace the "run" steps in  he build and deployment workflows in the original github actions workflow.  I guess the build and deployment jobs need parameters to be passed in to their respective build and deployment configurations as well.
4. I believe we could just retain the original logic in the build and deployment steps and just split the files into two separate files if that's the only thing that we are looking for and use the path trigger(which you seem to be using already) to kick off the relevant builds only.

simensma-fresh (approved):

Issue #779, opened 2022-12-08
Missing `schedule` object in the data sent to ODS


iman-jamali-fw:

iman-jamali-fw:
@T854236 

I just sent a telework submission on our DEV environment. Can you please check the `schedule` data on your end?

T854236:
@iman-jamali-fw 

I see 2 from this morning with missing values but the last one that you just sent has the schedule value.

GUID and IDIR information looks good too.

iman-jamali-fw:
@T854236 

The last one was after the fix was deployed to DEV.

@simensma-fresh I think we're good to deploy the fix to PROD. In the meantime, I'll do another test on TEST once it's deployed.

Stella-Archer:
YAY  - Push it to PROD please!

T854236:
@iman-jamali-fw 

yes confirming the schedule information is there.

iman-jamali-fw:
@T854236 

Can you please confirm if the schedule has been received by ODS on PROD? I just sent a submission after the fix is deployed to PROD

T854236:
@iman-jamali-fw 

Hi yes confirmed received.

PR #780, opened 2022-12-08
feature/DGJ_779-providing-a-list-of-object-keys-to-be-exclude-in-flatting-before-sent-to-ods

iman-jamali-fw:
## Summary

This PR will provide a way to exclude certain objects from faltting on send to ODS. #779 

## Changes
- Reading a list of excluded object keys in `SendSubmissionToODS` listener.
- excluding the objects that are in that list from flatting in `retrieveFormValues` method.


simensma-fresh (approved):
Nice!

Issue #706, opened 2022-11-09
TOOL: Formsflow pdf naming 


HasanSifat:
**User Story**
As a user (employee/manager) I want the downloaded pdf from the tool to be named in a way that it makes sense. Currently it downloads as form-exported-xxxxx.pdf

**Tasks**

- [ ] name the pdf same as the name of the draft 

**Acceptance Criteria**
When downloaded pdfs have the above formatted name

bhumin-fw:
@HasanSifat I guess you missed the suggested "PDF name" format in the task.

HasanSifat:
not sure what you are referring to @bhumin-fw

bhumin-fw:
**"add a feature so that the pdf name is downloaded as"**

Above line missed "PDF name" suggestion.
- @HasanSifat

HasanSifat:
@bhumin-fw can you send me a message in teams? Would be best to connect over a call on this

bhumin-fw:
@HasanSifat I do not have access to the BC Gov Team.

Attaching a screenshot if that shows the correct point, of what I'm asking. If not could you please add the meeting in teams, I'll get invited in EY email and can jump on a meeting.
![image.png](https://images.zenhubusercontent.com/62e98579ff9c075848ccdb67/f6c16faa-05a1-4358-9ce2-5da427d678a5)

HasanSifat:
my apologies lol, i added the criteria with closed brackets. Updated the ticket now and should be visible

iman-jamali-fw:
@Stella-Archer Please let me know if this format for the name of the pdf is what you're looking for:

- When the name field is populated in a form: `formName-Name`. For example: `Telework-iman_jamali`, ` Senior_Leadership_Review-hasan_sifat`.
- When the name filed is not populated: `formName-uniqueNumber`. For example: `Telework-1670304466059`

Stella-Archer:
Thanks @iman-jamali-fw , Yes formName-Name. For example: Telework-iman_jamali, Senior_Leadership_Review-hasan_sifat. make sense 
but I'm wondering if a name field is not populated instead of a random number we could amend it with current date 'Telework-2022-12-08' as I think if a user was downloading/saving a blank form they'd at least have reference to when the downloaded that version, which would be more helpful than a random number

iman-jamali-fw:
@Stella-Archer Thanks for reviewing the format. Sure, I'll replace the random number with the current DateTime.

[...]

PR #790, opened 2022-12-09
DGJ_706-TOOL-pdf-naming

iman-jamali-fw:
## Summary

This PR improves the TOOL's pdf naming by adding the formName and the name of the person filing the form to the exported pdf. #706 

## Changes
- Each form would send `formName` and `pdfName` to the TOOL.
- The PDF export function on the TOOL was updated to construct a proper name with the values it gets from the form
- Made sure the proper name was set when input values from the form is not sanitized or null.

### DIFF
--- a/forms-flow-web/src/components/Form/Item/Submission/Item/View.js
+++ b/forms-flow-web/src/components/Form/Item/Submission/Item/View.js
@@ -54,7 +54,7 @@ const View = React.memo((props) => {
         "printPdf",
         printToPDF
       );
-    }, 1000);
+    }, 2000);
     return () => {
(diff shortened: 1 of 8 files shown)


forms-flow-web/src/components/Form/Item/Submission/Item/View.js
```diff
@@ -54,7 +54,7 @@ const View = React.memo((props) => {
         "printPdf",
         printToPDF
       );
-    }, 1000);
+    }, 2000);
```
iman-jamali-fw, line 57: The interval was increased to make sure button become enabled.

simensma-fresh (approved):

PR #814, opened 2022-12-19
723 web api deployment

iman-jamali-fw:
## Summary

This PR complete the earlier work on web and webapi pipeline done by @bhumin-fw 

## Changes
- `web_bc.yaml` and `webapi_bc.yaml` were updated to only contain image stream
- workflow for both web and webapi were updated very similar to `forms-flow-documents`
- `Dockerfile.prod` path was updated to be picked up instead of `Dockerfile` with a path to that file. `forms-flow-documents` will need to be changed accordingly in another PR.
- All the GitHub actions were updated to the latest versions (e.g., checkout, docker cache, docker build and etc)
- webapi-dc file formatting (indentation) was updated to be consistent and error-prone.
- `Dockerfile.prod` for `forms-flow-web` get the images from Docker instead of `artifacts.developer.gov.bc.ca` since it doesn't work (required authentication otherwise) while the workflow runs in GitHub actions. @simensma-fresh Please let me know if this is fine.


bhumin-fw (approved):
A few changes are worth taking a look at that I didn't. Greta job Iman for the whole update here.

simensma-fresh (approved):
Awesome! 

Removing `artifacts.developer.gov.bc.ca/docker-remote` should be fine. That's just something you have to use within the OpenShift cluster when building to get around dockerhub's rate limiting

PR #838, opened 2023-01-06
DGJ-811 Hide content that has .hidden-in-print class and formio button

bhumin-fw:
## Summary

DGJ-811 Hide Content that has .hidden-in-print and formio button classes.

## Changes
- Set visibility for the floating and formio buttons.
- Set display property for the content that has class .hidden-in-print. If the content has no visibility, it will occupy space and put a blank page and block in PDF, hence hiding it.

## Notes
NA

### DIFF
--- a/forms-flow-web/src/services/PdfService.js
+++ b/forms-flow-web/src/services/PdfService.js
@@ -24,7 +24,13 @@ export const exportToPdf = ({formId, formName, pdfName}) => {
 export const printToPDF = ({ pdfName, formName }) => {
   // hide block with class .hidden-in-print
   const hiddenInPrint = document.querySelectorAll(".hidden-in-print");
-  hiddenInPrint.forEach((btmElm) => (btmElm.style.display = "none"));
+  let changeElm = [];
+  hiddenInPrint.forEach((elm) => {
+    if (elm.style.display === "" || elm.style.display === "block") {
+      elm.style.display = "none";
+      changeElm.push(elm);
+    }
+  });
   // Hiding the floating buttons during the PDF generation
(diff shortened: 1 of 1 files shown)


matbusby-fw (approved):
Nice!

forms-flow-web/src/services/PdfService.js
```diff
@@ -22,13 +22,18 @@ export const exportToPdf = ({formId, formName, pdfName}) => {
...
   // Hiding the floating buttons during the PDF generation
-  const floatingButtons = document.querySelectorAll("button.floatingButton");
+  const selectors = "button.floatingButton,.formio-component-button";
+  const floatingButtons = document.querySelectorAll(selectors);
   floatingButtons.forEach((btmElm) => (btmElm.style.visibility = "hidden"));
   toast.success("Downloading...");
   exportToPdf({ formId: "formview", pdfName, formName });
   setTimeout(() => {
     floatingButtons.forEach((btmElm) => (btmElm.style.visibility = "visible"));
+    hiddenInPrint.forEach((btmElm) => (btmElm.style.display = "block"));
```
iman-jamali-fw, line 36: Can we ensure these elements have had "block" for their display? If not, maybe you can save their display types before setting them to "none" and restore them here.

iman-jamali-fw (approved):
Looks great! Just had one comment.

forms-flow-web/src/services/PdfService.js
bhumin-fw: Yeah! that's a great catch. That must be the case.
I have updated the code with the condition. Hide visible complement, keep a track of them and make only those visible again.

PR #850, opened 2023-01-12
DGJ-845 No email if LDB user don't have bc.gov or bcldb email.

bhumin-fw:
## Summary

DGJ-845 No email if LDB user don't have gov.bc.ca or bcldb.com email.

## Changes
Check new field in case of LDB preferredEmail. If it includes gov.bc.ca or bcldb.com domain then we will continue as old one. If not then we will show different detail in success page and not send PDF attachment in email (this one will handle in camunda).

## Notes
NA


iman-jamali-fw (approved):
LGTM!

Issue #848, opened 2023-01-11
PDF generation timeout issue


iman-jamali-fw:

iman-jamali-fw:
There is an issue when a pdf generated in `forms-flow-documents` takes longer (more than ~5 seconds), Camunda workflow fails with a timeout issue.

Increasing `forms-flow-documents` indirectly improves this issue as it helps the time of pdf generation decreases.

A potential solution would be to increase the Camunda HTTP client timeout.

iman-jamali-fw:
Successfully tested on TEST and DEV. Ready to be moved to PROD.

PR #853, opened 2023-01-12
DGJ_848-pdf_generation_timeout_issue

iman-jamali-fw:
## Summary

This PR fixes the Camunda timeout issue when a task takes more than 5 seconds. The issue showed up sometimes during the PDF generation. #848 

## Changes
- The default timeout time was increased from 5 to 30 seconds
- A env was added to be able to modify this value in the future with ease


matbusby-fw (approved):
🎉

bhumin-fw (approved):
Looks good to me.

PR #864, opened 2023-01-19
hotfix/camunda-timeout-increase

iman-jamali-fw:
## Summary

On PROD, if multiple pdfs are generated simultaneously, Camunda still fails with a timeout issue. To fix that, we need to increase the timeout from the 30s to the 80s. For some reason the env for that timeout is not working, so increasing it in the code.

## Changes
- timeout increased to 80s.


matbusby-fw (approved):
Sounds good to me!

bhumin-fw (approved):

PR #891, opened 2023-02-07
DGJ_888-extract_bceid_last_name

iman-jamali-fw:
## Summary

This PR fixes the issue with BCeID the last name being automatically extracted from `display_name`.

Context: BCeID only provides `display_name` and not providing first_name and last_name separately. `display_name` is passed as the first_name through the Keycloak. To fix that, as a workaround employee.me endpoint split them and pass them separately to the FE.

## Changes
- Extracting first_name and last_name from BCeID display_name.
- Also updated employeeData models to not break if either first_name or last_name is null during the creation of display_name.

## Notes
Another issue that was fixed as part of this PR, through Keycloak Configs, was the initial BCeID prompt to update their info including first and last names. To force Keycloak to not to show that form (listed in the image below), a new login flow was assigned to BCeID identity Provider (listed in the image below)

Screenshot 2023-02-02 at 11 03 41 AM

Screenshot 2023-02-06 at 8 28 35 PM

### DIFF
--- a/forms-flow-api/src/formsflow_api/models/employee_data.py
+++ b/forms-flow-api/src/formsflow_api/models/employee_data.py
@@ -5,7 +5,7 @@ class EmployeeData():
   # data: Response retrieved from the ODS
   # Spec of possible values can be found at: https://analytics-testapi.psa.gov.bc.ca/apiserver/api.rst#Datamart_Telework_employee_demo
   def __init__(self, data):
-   self.displayName = data["first_name"] + " " + data["last_name"]
+   self.displayName = " ".join(filter(None, (data["first_name"], data["last_name"])))
    self.name = data["name"]
    self.firstName = data["first_name"]
    self.lastName = data["last_name"]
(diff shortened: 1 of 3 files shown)


forms-flow-api/src/formsflow_api/models/employee_data.py
```diff
@@ -5,7 +5,7 @@ class EmployeeData():
   # data: Response retrieved from the ODS
   # Spec of possible values can be found at: https://analytics-testapi.psa.gov.bc.ca/apiserver/api.rst#Datamart_Telework_employee_demo
   def __init__(self, data):
-   self.displayName = data["first_name"] + " " + data["last_name"]
+   self.displayName = " ".join(filter(None, (data["first_name"], data["last_name"])))
```
iman-jamali-fw, line 8: This would prevent the whole endpoint to break if either first_name or last_name was null.

bhumin-fw (approved):
It looks good! :)

matbusby-fw (approved):
Looks good!

PR #902, opened 2023-02-08
DGJ-883 Form page loading issue.

bhumin-fw:
## Summary
When user redirect to form page after loading, It load a blank page.

## Changes
It was fail due to user object was null and that is not checked.
User object load several time if user land on form edit page directly. In couple attempts user object will have value.
Only check "user" along with STAFF_DESIGNER condition lead to "component re-render failed due to previous render not done".
To handle it, call useEffect on user object change and not under user object not null.

## Notes
I have tested it will basic cases. But this needs to test with the actual user from the different provider like idir, BCeID, BCSC and STAFF_DESIGNER.

### DIFF
--- a/forms-flow-web/src/components/Form/Item/View.js
+++ b/forms-flow-web/src/components/Form/Item/View.js
@@ -423,7 +423,7 @@ const View = React.memo((props) => {
         }
       }
     }
-  }, [user]);
+  });
 
   if (isActive || isPublicStatusLoading || formStatusLoading) {
     return (
--- a/forms-flow-web/src/components/Form/Item/View.js
+++ b/forms-flow-web/src/components/Form/Item/View.js
@@ -118,7 +118,7 @@ const View = React.memo((props) => {
   const [notified, setNotified] = useState(false);
   const [defaultVals, setDefaultVals] = useState({});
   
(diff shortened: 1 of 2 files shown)


iman-jamali-fw (approved):
Looks good to me.

forms-flow-web/src/components/Form/Item/View.js
```diff
@@ -409,25 +409,21 @@ const View = React.memo((props) => {
...
         }
-      }, 1000);
-      return () => {
-        clearInterval(formAccessInterval);
-      };
-    });
-  }
+      }
+    }
+  });
```
iman-jamali-fw, line 426: I removed the `[user]` to make sure check for `formRef.currect` runs until it is available.

iman-jamali-fw:
For the Google user on Local. I need to do two things. First, make sure username of Google users have postfix of `_google` which I'll do today for Local Keycloak. And you need to add `google` to the list of supported IDPs in each form you want to user google user in.

For ease of testing, you can easily connect your local to DEV's Keycloak, which is what I did to test out this branch. I can give you the .env if something you wanted to do in the future.

bhumin-fw:
Thanks Iman for the details and updates. I did use the same trick to check google.
And yea, Please share .env that help me to connect with dev keycloak from local.

PR #937, opened 2023-02-17
Revert peoplesoft latest data and ignore UNMOUNT event on draft page

bhumin-fw:
## Summary

Draft saved with previous ID on an unmount event.

## Changes
- Stop saved on unmount event on draft edit page
- Revert fetching peoplesoft data on draft page

## Notes
Issue seems legacy, draftSubmission has latest value, but for UNMOUNT event it fire with previous values.


matbusby-fw (approved):
Looks good to me!

Issue #908, opened 2023-02-09
1.10: Popup for alternate dispute resolution info 


fazil-ey:
Create a popup with the following content to provide additional context to people in 1.10 form. 
The popup should be similar to the release note popup on digital journeys portal

This popup should be an (i) tooltip next to this question - "Are you open to participating in an alternate dispute resolution process, rather than an investigation?" 

> Alternative dispute resolution – conducted through the Conflict Management Office in PSA, a facilitator can help resolve interpersonal workplace conflict. Participation is voluntary and based on an informal, interest-based approach.
> 
> Investigation – conducted under the Public Service Employment Act, an investigator will interview parties relative to a complaint, collect and assess evidence, and determine findings of fact needed to resolve the complaint.
> 
> Review – any actions outside of an investigation in which the appropriate manager takes steps, with the support of PSA staff as appropriate, to resolve the issues and incidents outlined in the complaint.
> 

Refer - #673
Example for how the pop up should look like 
![Screen Shot 2023-02-09 at 1.35.13 AM.png](https://images.zenhubusercontent.com/62e986a3ff9c07df27cce462/a345db0a-78d0-4dc1-9cf1-69f1470c9e7f)

iman-jamali-fw:
@fazil-ey 
When do you want this popup to show up? Is that displayed when a new form is opened or it triggers by a button within the form?

MeghanStothers:
Waiting for Fazil/Stella to see if this is what is wanted

fazil-ey:
@iman-jamali-fw this looks good. Just want to ensure that for 1.10 we either want an (i) like we use for tooltips or a link to trigger the popup

MeghanStothers:
ready to test

iman-jamali-fw:
@Stella-Archer  @Andrew-Vargas-bcgov 

This is ready to be tested on TEST. Documentation was also added [here](https://github.com/bcgov/digital-journeys/blob/main/docs/forms.md#form-access)

Andrew-Vargas-bcgov:
@iman-jamali-fw Tested and works :).

MeghanStothers:
approved and ready to move to deployment

Stella-Archer:
A new term "recipient'  has been added to the definitions:

Recipient – the person who will receive the complaint, normally the supervisor or excluded manager of the person submitting the complaint. They are not a subject of the complaint but they will be involved in helping to resolve the complaint.

Stella-Archer:
[...]

Issue #907, opened 2023-02-09
1.10 : Contact Us Button


fazil-ey:
Add a contact us button next to save and download pdf buttons in the 1.10 form so that applicants can easily know who they can contact to get help as they fill the forms

The contact us button should open a popup with the following content 

> •	For help completing the form for BCGEU members, contact your GEU at 1-800-663-1674
•	For help completing the form for PEA members, contact your PEA at 1-800-779-7736
•	For technical help in using the form, contact digitaljourneys@gov.bc.ca
•	For PSA accommodation assistance in completing the form, contact AskMyHR 
•	For LDB accommodation assistance in completing the form, contact employee.relations@bcldb.com 
•	For suggestions on how to improve the form, contact PSA.RegistrarIntake@gov.bc.ca

Example for reference on how the popup should look like  ![Screen Shot 2023-02-09 at 1.35.13 AM.png](https://images.zenhubusercontent.com/62e986a3ff9c07df27cce462/4ac8934b-a547-4b27-b393-5e2282759f8c)

MeghanStothers:
Component created - Iman looking into options for tool tip to trigger the pop up

MeghanStothers:
Ready to test

iman-jamali-fw:
@Stella-Archer  @Andrew-Vargas-bcgov 

This is ready to be tested on TEST. Documentation was also added [here](https://github.com/bcgov/digital-journeys/blob/main/docs/forms.md#form-access)

Andrew-Vargas-bcgov:
Tested and works perfectly in DEV. Great work @iman-jamali-fw 
In DEV I added mailto links for the emails and a link to myhr which I think we had left out of the ticket.

iman-jamali-fw:
Perfect! Thanks @Andrew-Vargas-bcgov 

Could you please add that link to TEST as well? 

Forms on TEST are the most important ones, during the deployment to PROD, we basically copy the form on TEST and paste it onto PROD.

To optimize your testing efforts, you can only test the features on TEST.

Andrew-Vargas-bcgov:
Thanks @iman-jamali-fw I have made the changes in TEST as well.

MeghanStothers:
Andrew tested and looks okay. Can close this ticket

PR #943, opened 2023-02-28
DGJ_908-info-modal-component

iman-jamali-fw:
## Summary
This PR introduces a reusable popup component that can be triggered by a custom button in the forms. In addition, textField and textArea components in the Formio were extended to support a tooltip that can trigger that popup component.

The popup component can receive `title` in text and `body` in text/HTML format.

This PR completes two tickets #907  and #908 

## Changes
- View and Draft pages now support a reusable modal component that can be triggered by a custom event
- `textField` and `textArea` components were extended to enable tooltips to open in the popup by passing a special HTML comment in the first line of the tooltip text. `<!-- popup -->`

Screenshot 2023-02-28 at 1 55 31 PM

### DIFF
--- a/forms-flow-web/src/containers/MessageModal.jsx
+++ b/forms-flow-web/src/containers/MessageModal.jsx
@@ -12,7 +12,9 @@ const MessageModal = React.memo((props) => {
         <Modal.Header>
           <Modal.Title>{title}</Modal.Title>
         </Modal.Header>
-        <Modal.Body>{message}</Modal.Body>
+        <Modal.Body>
+          <div dangerouslySetInnerHTML={{ __html: message }} />
+        </Modal.Body>
         <Modal.Footer>
           <Button type="button" className="btn btn-default" onClick={onConfirm}>
             {t("Ok")}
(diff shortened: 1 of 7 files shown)


forms-flow-web/src/containers/MessageModal.jsx
```diff
@@ -9,10 +9,12 @@ const MessageModal = React.memo((props) => {
   return (
     <>
       <Modal show={modalOpen}>
```
bhumin-fw, line 11: do you think we should add `onHide={onConfirm}` here? It will allow user to close popup by click anywhere on backdrop and do not need to reach to "ok" button.

bhumin-fw (approved):
Looks good to me! 👍 
Only one comment, Added one suggestion regarding popup close.
Nice idea to process details, it is very easy to extend same custom event for any component.

forms-flow-web/src/containers/MessageModal.jsx
iman-jamali-fw: That's a good one. I'll add it.

PR #944, opened 2023-03-01
hotfix/DGJ_908-info-modal-component

iman-jamali-fw:
## Summary

This Hotfix fixes the issue with tooltip info buttons supporting popups. Sometimes their previous event listeners were triggered which means a tooltip shows the tooltip when hovered as opposed to only showing the popup when clicked.

## Changes
- Popup tooltips were cloned and replaced to drop all the previous eventListneres

Screenshot 2023-03-01 at 9 41 05 AM

Screenshot 2023-03-01 at 9 43 31 AM


matbusby-fw (approved):
Ah, nice catch!

PR #957, opened 2023-03-04
Fix form access issue on draft edit page.

bhumin-fw:
## Summary

Fix form access issue on draft edit page,
correct redirection to form…

## Changes
- apply same logic as view.js.
If user is designer, allow access. If form meta supportedIdp set then check for the form access and if it is blank skip form access check.
- Redirect back to draft list page instead of form page.

## Notes
N/A


iman-jamali-fw (approved):
LGTM

PR #967, opened 2023-03-08
DGJ-841 added custom CSS to adjust UI

bhumin-fw:
## Summary
In data grid, We have to adjust few button and component padding

## Changes
- Added 2 custom class and CSS. If we add those custom class then and then css will apply. So it will not affect any existing container, layout or grid components.

## Notes
NA


iman-jamali-fw (approved):
LGTM

PR #972, opened 2023-03-10
DGJ-964 an API to add user in Keycloak and assign group

bhumin-fw:
## Summary
- Added an API to automation add new users and assign provided group. 
- If user is already exist with same email/idp then only assign group.

## Changes
- Added webapi that accept user(first & last name, email and idp) object as array and group name.
- API will return same user object response with additional details like "user_id", "success" and "updated" with group to help issuer(developer or designer user) save response or display(If we integrate it with frontend).
- Only "Form Designer" are able to call this API.

## Notes
I'll add documentation and sample data for the API. As it will be handle by developer only(unless we add frontend for the same)
It will help to trigger the API from local script with designer token or from postman app. It will not required developer to collect keycloak configuration as those are already there in the WEBAPI.

### DIFF
--- a/forms-flow-api/src/formsflow_api/resources/employeeData.py
+++ b/forms-flow-api/src/formsflow_api/resources/employeeData.py
@@ -3,9 +3,14 @@
 from http import HTTPStatus
 from flask import g, current_app, request
 from flask_restx import Namespace, Resource
-from formsflow_api_utils.utils import auth, cors_preflight, profiletime
+from formsflow_api_utils.utils import (
+    DESIGNER_GROUP,
+    auth,
+    cors_preflight,
+    profiletime,
+)
 from formsflow_api_utils.exceptions import BusinessException
-from formsflow_api.services import EmployeeDataService
+from formsflow_api.services import EmployeeDataService, KeycloakService
 
 
 
@@ -65,3 +70,31 @@ class EmployeeNames(Resource):
             return err.error, err.status_code
 
         return employee_names, HTTPStatus.OK
+
+
+@cors_preflight("POST, OPTIONS")
+@API.route("/user-group", methods=["POST", "OPTIONS"])
+class EmployeeGroup(Resource):
+    """Employee and Group Data"""
+
+    @staticmethod
+    @profiletime
+    @auth.require
+    @auth.has_one_of_roles([DESIGNER_GROUP])
+    def post():
+        """Process users"""
+        try:
+            data = request.get_json()
+            group = data.get("group")
+            users = data.get("users")
+            if group and users:
+                obj = KeycloakService()
+                result, groupId = obj.add_users(group, users)
+                if groupId:
+                    # print(er)
+                    return result, HTTPStatus.OK
+                return {"errorMessage": "Group not found"}, HTTPStatus.NOT_FOUND
+            else:
+                return {"errorMessage": "Provide group and users"}, HTTPStatus.BAD_REQUEST
+        except Exception as e:
+            return {"errorMessage": f"Resource function error: {e}"}, HTTPStatus.BAD_REQUEST
--- a/forms-flow-api/src/formsflow_api/resources/employeeData.py
+++ b/forms-flow-api/src/formsflow_api/resources/employeeData.py
@@ -88,10 +88,9 @@ class EmployeeGroup(Resource):
             group = data.get("group")
             users = data.get("users")
             if group and users:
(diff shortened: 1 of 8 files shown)


forms-flow-api/src/formsflow_api/resources/employeeData.py
```diff
@@ -65,3 +70,31 @@ def get():
...
+        """Process users"""
+        try:
+            data = request.get_json()
+            group = data.get("group")
+            users = data.get("users")
+            if group and users:
+                obj = KeycloakService()
+                result, groupId = obj.add_users(group, users)
+                if groupId:
+                    # print(er)
```
matbusby-fw, line 94: Can probably remove this commented out print statement

matbusby-fw (approved):
Looks great Bhumin!

forms-flow-api/src/formsflow_api/resources/employeeData.py
```diff
@@ -65,3 +70,31 @@ def get():
...
+    @auth.require
+    @auth.has_one_of_roles([DESIGNER_GROUP])
+    def post():
+        """Process users"""
+        try:
+            data = request.get_json()
+            group = data.get("group")
+            users = data.get("users")
+            if group and users:
+                obj = KeycloakService()
```
iman-jamali-fw, line 91:
```suggestion
                kcAdminClient = KeycloakService()
```

forms-flow-api/src/formsflow_api/services/keycloak_service.py
```diff
@@ -0,0 +1,97 @@
...
+            result["message"] = f"{fields} fields required with valid data"
+            return result
+
+        username = userData.get("email")
+        idp = userData.get("idp")
+        if idp and idp !="":
+            username = f"{username}_{idp}"
+        userData["username"] = username
+        userData["enabled"] = True
+        userId = self.get_user_id(username)
```
iman-jamali-fw, line 52: Here, you should get the user id by email instead of the username because you don't know what the username will look like for that user.

username for IDIR: `${email}+_idir`
username for BCeID: `${bceid_username}_bceid`
username for BCSC: `${bcsc_guid}_bcsc`

So, until BCeID and BCSC users not logged in we don't know what their usernames would look like. But we know the email and the IDP. Therefore your `get_user_id` logic should be like this:
```
1. find users by email >> can return more than one users
2. Look for the user whose username contains that IDP postfix. For example, if IDP is BCeID, you should filter users whose username has a postfix of "_bceid".
3. return the id that user
```

forms-flow-api/src/formsflow_api/services/keycloak_service.py
```diff
@@ -0,0 +1,97 @@
...
+            userData = RequestUserSchema().load(userData)
+        except ValidationError as e:
+            fields = ",".join(e.messages.keys())
+            result["message"] = f"{fields} fields required with valid data"
+            return result
+
+        username = userData.get("email")
+        idp = userData.get("idp")
+        if idp and idp !="":
+            username = f"{username}_{idp}"
```
iman-jamali-fw, line 49: This only works for `IDIR` users. for non-idir users at this point, Keycloak only merges based on the username and we don't know their username until they log in.

A workaround would be to find out how to make Keycloak merge accounts based on the email in each IDP.

iman-jamali-fw: This is fine for now because for the Senior Leader form, we only need IDIRs. We can improve this later.

iman-jamali-fw: Please limit this feature for IDIR at the moment until we figure out how to scale it for other IDPs.

forms-flow-api/src/formsflow_api/resources/employeeData.py
bhumin-fw: Fixed

forms-flow-api/src/formsflow_api/resources/employeeData.py
bhumin-fw: Updated variable name

forms-flow-api/src/formsflow_api/services/keycloak_service.py
bhumin-fw: limited this feature for IDIR only

iman-jamali-fw (approved):
LGTM

PR #1009, opened 2023-03-27
fix/DGJ_558-delete-incomplete-submission

iman-jamali-fw:
## Summary

This pull request enhances the submission/application deletion listener in Camunda by enabling it to handle situations where the relevant data cannot be located. For instance, if a submission has already been deleted due to any reason, the delete listener will not trigger an error in Camunda. Instead, it will proceed with deleting the application and advancing the workflow seamlessly.


bhumin-fw (approved):
Looks good to me!

PR #1032, opened 2023-04-11
DGJ-1000 and DGj-990 API to fetch ministry, city and worksite for form

bhumin-fw:
## Summary
Added APIs to fetch ministries, cities and worksites.

## Changes
Add API in form-flow-webapi, and data fetches from ODS API endpoints.

## Notes
N/A

### DIFF
--- a/forms-flow-api/src/formsflow_api/services/influenza_service.py
+++ b/forms-flow-api/src/formsflow_api/services/influenza_service.py
@@ -14,7 +14,7 @@ class InfluenzaService:
                         headers={"Authorization": current_app.config.get("ODS_AUTH_TOKEN")})
         except Exception as e:
             raise BusinessException(
-                {"message": "Failed to look up user in ODS"}, HTTPStatus.INTERNAL_SERVER_ERROR
+                {"message": "Failed to look up ministry in ODS"}, HTTPStatus.INTERNAL_SERVER_ERROR
             )
         
         ministry_names_res = response_from_ods.json()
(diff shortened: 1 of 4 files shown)


forms-flow-api/src/formsflow_api/services/influenza_service.py
```diff
@@ -0,0 +1,60 @@
...
+    @staticmethod
+    def get_ministry_names():
+        """fetch ministry_names"""
+        try:
+            ministry_api_url = "{}/ODS_Datamart_influenza_ministry".format(current_app.config.get("ODS_URL"))
+            response_from_ods = requests.get(ministry_api_url,
+                        headers={"Authorization": current_app.config.get("ODS_AUTH_TOKEN")})
+        except Exception as e:
+            current_app.logger.error(e)
+            return []
```
iman-jamali-fw, line 16: Returns `[]` in case of an exception might confuse the client. Instead, let's raise an error and catch it in the resource class, so we can send a more helpful message to the client. Check out the EmployeeData resource and service files for examples of how to do this.

iman-jamali-fw (changes requested):
All looks good. Please improve the error handling a little bit :)

forms-flow-api/src/formsflow_api/services/influenza_service.py
bhumin-fw: @iman-jamali-fw 
`[]` set intentionally. as in response If other than 200 code it stopped working in form Select Component.

iman-jamali-fw: Can you move that logic to the client side, in this case, the Select component to handle the error?

matbusby-fw (approved):
Looks solid to me!

iman-jamali-fw (approved):
Looks great! Thanks for improving the error handling 👍🏼

PR #1037, opened 2023-04-13
DGJ-1024 workflow supporting API endpoints and data transform

bhumin-fw:
## Summary

- Added code to convert camelCase to snake_case in send_to_ods delegates on BPM.
- Added supported method for cold/flu workflow

## Changes
- Added case conversation
- Update API endpoint details to fetch data from ODS
- Add new GET registration API endpoint
- Remove city API

## Notes
N/A

### DIFF
--- a/forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/SendSubmissionToODSDelegate.java
+++ b/forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/SendSubmissionToODSDelegate.java
@@ -14,6 +14,7 @@ import org.camunda.bpm.extension.commons.connector.HTTPServiceInvoker;
 import javax.inject.Named;
 import java.io.IOException;
 import java.util.Map;
+import java.util.HashMap;
 import java.util.List;
 import java.util.ArrayList;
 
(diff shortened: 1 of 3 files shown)


forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/SendSubmissionToODSDelegate.java
```diff
@@ -82,9 +84,14 @@ private void sendSubmissionToODS(DelegateExecution execution) throws IOException
         if (managerGuid != null) {
             values.put("manager_guid", String.valueOf(managerGuid));
         }
-
         ObjectMapper objectMapper = new ObjectMapper();
-        String json = objectMapper.writeValueAsString(values);
+        String json = "";
+        if (objectKeycase.equals("snake_case")) {
```
matbusby-fw, line 89: I'm not super familiar with these functions, but isn't this saying "if `objectKeycase` is `snake_case` then convert it to `snake_case`?

matbusby-fw: Oh, is the `objectKeycase` the desired form for the variable?

matbusby-fw (approved):
Looks good to me I think (just wasn't sure about how `objectKeycase` was implemented, but I think I get it now.

iman-jamali-fw (approved):

PR #1049, opened 2023-04-14
employeeData service logs were removed, employeeData model was imporo…

iman-jamali-fw:
## Summary
This PR improve the `employee_data` model to gracefully handle the case where a piece of data was missing from ODS. Previously, if one part of employee data is missing, it broke and returned no value for that employee.

Also removed the debugging logs.


matbusby-fw (approved):
Nice enhancement!

PR #1050, opened 2023-04-14
fix/DGJ_884-employee-info-query-builder-missing-&

iman-jamali-fw:
## Summary

This PR adds a missing "&" in the get employee info query to ODS.


matbusby-fw (approved):
👍

PR #1052, opened 2023-04-17
fix/DGJ_884_get-info-error-handling-and-set-component-value-improvments

iman-jamali-fw:
## Summary

This PR improves the error handling of the employee get info endpoint which previously always returned 500 errors. Also improves the FE set the value to formio component to do a deep iteration over the formio components, now can find the nested components in the wizard form.


matbusby-fw (approved):
🌮

Issue #1025, opened 2023-04-04
Cold/Flu: Delete form submission - Logic


bhumin-fw:
As a user I want to delete a submitted form request.

Delete form submission required to handle separately, as Delete needs to enable for selected Forms only. 
In Cold/Flu form, data needs to soft delete from data warehouse.
To do so, Delete submission needs to handle separately, it will remove form entry and send DELETE request to data warehouse APIs.

Acceptance Criteria:
- Delete submitted form
- Soft delete data on data ware house by triggering DELETE api endpoint.

iman-jamali-fw:
@ca-liu @fazil-ey @bhumin-fw 

We currently don't have to delete the submitted form functionality. However, there is similar functionality for drafts.![Screenshot 2023-04-25 at 11.54.00 AM.png](https://images.zenhubusercontent.com/61ddc48547f476ed3f35ca28/1a09bd60-d68a-41a6-a69b-b2eed06ba180)![Screenshot 2023-04-25 at 11.54.05 AM.png](https://images.zenhubusercontent.com/61ddc48547f476ed3f35ca28/354768c8-bef9-45fa-9530-20136bd993cb)


This ticket only covers the logic side of the work that includes
- Deletion from DGJ DBs
- Delete from the warehouse

There is another piece of work, **the UI side** that should clear how the delete submitted form functionality is added, and also clarify the compatibility of deleted submitted forms with other forms (SL Review, Mat/Pat). Do we want the delete only to be enabled for Cold/Flu submissions or for all? 

I'd suggest another ticket for UI and use this ticket to track the logic.

iman-jamali-fw:
related Mike's comments from #1000 are to be continued in this ticket:

> Endpoint for registration soft deletion is now available:
> HTTP POST (https://analytics-testapi.psa.gov.bc.ca/apiserver/api.rsc/ods_datamart_influenza_delete_registration/)
> 
> var postdata = { "application_id": 1};
> 

> Yes, the "deleted" registrations will be ignored from endpoint /ods_datamart_influenza_vw_registrations_worksites_joined. As for endpoint /ods_datamart_influenza_registrations/, because it is directly with table, you need to apply filter "deleted_ind Eq null"

ca-liu:
> @ca-liu @fazil-ey @bhumin-fw 
> 
[...]

PR #1074, opened 2023-04-26
feature/DGJ_1025-delete-cold-flu-registration-logic

iman-jamali-fw:
## Summary

This PR allows users to delete the submitted applications -- a requirement for Cold/flu submissions. This is the logic part of #1025 ticket

## Changes

- Added an endpoint to delete submissions from the ODS warehouse
- Added/updated delete application from webApi/formio DBs from an endpoint which was added a while ago but left with no use ♻️

## Notes

At this point, regular users can only delete their own applications and a designer can delete their own and others. That part of the logic will be updated when a new group/role for cold/flu admin is added to handle that special condition.

### DIFF
--- a/forms-flow-api/src/formsflow_api/resources/influenzaWorksite.py
+++ b/forms-flow-api/src/formsflow_api/resources/influenzaWorksite.py
@@ -124,3 +124,20 @@ class InfluenzaWorksite(Resource):
         except BusinessException as err:
             current_app.logger.warning(err.error)
             return err.error, err.status_code
+
+@cors_preflight("DELETE, OPTIONS")
+@API.route("/worksites_registrations/<int:application_id>", methods=["DELETE", "OPTIONS"])
+class InfluenzaWorksiteById(Resource):
+    """Resource for worksites_registrations by id"""
+
+    @staticmethod
+    @profiletime
+    @auth.require
+    def delete(application_id: int):
+        """delete influenza worksites_registrations by application_id"""
+        try:
+            InfluenzaService.delete_worksites_registrations(application_id)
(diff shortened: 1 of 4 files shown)


forms-flow-api/src/formsflow_api/resources/influenzaWorksite.py
```diff
@@ -124,3 +124,21 @@ def get():
...
+    @auth.require
+    @auth.has_one_of_roles(["formsflow-designer"])
+    def delete(application_id: int):
+        """delete influenza worksites_registrations by application_id"""
+        try:
+            InfluenzaService.delete_worksites_registrations(application_id)
+            return "Deleted", HTTPStatus.OK
+        except BusinessException as err:
+            current_app.logger.warning(err.error)
+            return err.error, err.status_code
```
iman-jamali-fw, line 144: Enabling a designer to directly remove an application in ODS warehouse. Not an exact requirement, added just in case 🙃

bhumin-fw (approved):
good point that you covered!

forms-flow-api/src/formsflow_api/resources/influenzaWorksite.py
bhumin-fw: @iman-jamali-fw user will be able to delete their own form. does this decorator allow that too?

iman-jamali-fw: Not from this endpoint. 

There is another endpoint in the `application` resource that will allow the users to delete their own submissions (remove it from all related data sources). This one is just in case the designer wants to delete it from ods directly.

PR #1121, opened 2023-05-09
DGJ_1082-cold-flu-admin-add-to-formio-role-token

iman-jamali-fw:
## Summary

This PR adds cold-flu-admin role to the formio token.


bhumin-fw (approved):
looks good!

Issue #1082, opened 2023-05-01
Cold/Flu: Admin can EDIT all cold/flu submissions


fazil-ey:
**User Story**
As a Cold/Flu admin, i should be able to edit any cold/flu submission by any registrant, so that I am able to make any necessary changes/corrections on the form without depending on the registrant  

**AC**
- [x] Admin should be able to view all cold/flu submissions
- [x] Admin should be able to edit any field on the form for all cold/flu submissions. On submission the form must reflect the edited values. 
- [x] Admin should be NOT be able to view submissions by other users for any other form

ayushdamani:
Tested with @iman-jamali-fw !
Some cache issue are not allowing the Admin to edit and save the form.

iman-jamali-fw:
@ayushdamani 

The issue if fixed and we can test it again.

ayushdamani:
Tested and verified working. As an Admin, I was able to:
- view all cold/flu submissions
-  edit any field on the form for all cold/flu submissions.

Acceptance criteria met ✔️

Groups selected in Keycloak - 
![Screenshot 2023-05-11 113921.png](https://images.zenhubusercontent.com/63f54aecd19de9f18f88a815/b49ebce0-56d1-4414-a331-67c0811e35ec)

MeghanStothers:
Not approved - Heather needs to able to edit this filed too: 
![image.png](https://images.zenhubusercontent.com/61ddb02eac135d772cfce6ec/694b4a57-0487-45fc-9560-c910777f315e)

Stella-Archer:
works terrific! Thx. PO Approved. Tix closed

Issue #1075, opened 2023-04-27
Cold/Flu: Configure cold/flu admin group in Keycloak


bhumin-fw:
As a user I want to add new group name "cold-flu-admin" into the Keycloak.

Task: 

- [ ] Add new group with name "cold-flu-admin" in Keycloak.
- [ ] Add user (existing or new) to that group.

fazil-ey:
Additional notes from Stella 

Admin functionality
All submitted forms should show with the person’s name. In the absence of “editing” capability by the user, Admin need to be able to “search” by name when editing submissions.

Need to confirm whether ADMIN will have the ability to view and EXPORT all data to excel (via button), vs. needing to schedule report runs and/or emailing a person on dev team to run and send it (Caleb exploring with Don?)
NO email receipt to user when submission is edited

ADMIN (Cold.Flu@gov.bc.ca) to receive email receipt of any NEW or EDITED submissions (AFTER the close date)… No notification during the registration period.

iman-jamali-fw:
@ayushdamani 

We're set for testing on TEST! To make a cold-flu-admin user, please just add them to both client and cold-flu-admin groups in Keycloak. If you need a hand, feel free to ask me! With this setup, the cold/flu admin user can see all cold/flu forms submitted by others and their own submissions for any form type.

Stella-Archer:
@chriscaldwell-psa could you help me add Jodi Little to the keycloak groups?

iman-jamali-fw:
@ayushdamani 

The issue is fixed and it's ready for testing on TEST.

ayushdamani:
Thank you, Iman!

Tested and verified working. Acceptance criteria met ✔️

A shot from keycloak - 
![Screenshot 2023-05-11 113921.png](https://images.zenhubusercontent.com/63f54aecd19de9f18f88a815/a1b8941a-e5af-4969-b3d1-668f096e298e)

MeghanStothers:
Tested and approved with Stella and Heather

PR #1123, opened 2023-05-11
fix/DGJ_1082-cold-flu-admin-role-on-form-update

iman-jamali-fw:
## Summary

This PR fixes the issue of dropping `cold-flu-admin` submission access during the form update in the form builder. #1082 #1075 

## Changes

- Check to see if the `cold-flu-admin` role is part of the editing form, and keep it in that case


bhumin-fw (approved):
This looks good! Thanks Iman, this is a huge addition in system that we will be able to leverage in new forms.

PR #1133, opened 2023-05-15
DGJ_1117-application-last-update-missing-on-edit

iman-jamali-fw:
## Summary

This PR Adds a PATCH endpoint to webApi that allows updating the `modefied` field. 

## Changes
- PATCH `application/id` endpoint was added
- application update service was updated to add `modified` date to the list of supported update fields

## Notes
 Each form that changes after initial submission through website would require to hit the update endpoint to make sure its update time is reflected in the respective application. We can add the below snippets to the button that update a submission on each form:

```js
// Call Web API to update the application
      const webApiBaseUrl = window._env_?.REACT_APP_WEB_BASE_URL || "http://localhost:5001";
      const applicationId = data.applicationId;
      const url = `${webApiBaseUrl}/application/${applicationId}`;
      const headers = new Headers({
        Authorization: 'bearer ' +  localStorage.getItem("authToken"),
        'Content-Type': 'application/json'
      });

      fetch(url, {headers, method: "PATCH"})
        .then(response => {
          if (!response.ok) {
            console.error(response);
          } else {
            return response.json();
          }
        })
        .then(fetchData => {})
        .then(error => {
          console.error(error);
        });
```

### DIFF
--- a/forms-flow-api/src/formsflow_api/resources/application.py
+++ b/forms-flow-api/src/formsflow_api/resources/application.py
@@ -507,35 +507,3 @@ class ApplicationResourceByIdDelete(Resource):
         except BusinessException as err:
             current_app.logger.error(err.error)
             return err.error, err.status_code
-
-# @cors_preflight("PATCH, OPTIONS")
-# @API.route("/<int:application_id>/update", methods=["PATCH", "OPTIONS"])
-# class ApplicationResourceByIdUpdate(Resource):
-#     """Update application date by id."""
-
-#     @staticmethod
-#     @auth.require
-#     @profiletime
-#     @user_context
-#     def patch(application_id, **kwargs):
-#         """Update application date by id."""
-#         try:
-#             data = {"modified": datetime.datetime.now()}
-#             ApplicationService.update_application(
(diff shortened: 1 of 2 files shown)


forms-flow-api/src/formsflow_api/resources/application.py
```diff
@@ -470,4 +508,36 @@ def delete(application_id, **kwargs):
             return f"Application was successfully deleted from all DB sources with id: {application_id}", HTTPStatus.OK
         except BusinessException as err:
             current_app.logger.error(err.error)
-            return err.error, err.status_code
\ No newline at end of file
+            return err.error, err.status_code
+
+# @cors_preflight("PATCH, OPTIONS")
+# @API.route("/<int:application_id>/update", methods=["PATCH", "OPTIONS"])
```
bhumin-fw, line 512: is this method used anywhere in DJ system?

iman-jamali-fw: Oops! This was left here by mistake. Removed

bhumin-fw (approved):
It looks good Iman!

Issue #1100, opened 2023-05-04
Mat/Pat - CRM ticket on approval


fazil-ey:
**User Story**
As a supervisor, I want a CRM ticket to be created automatically when I approve a mat/pat leave application so that I do not have to do it manually

**Acceptance Criteria**
- [ ] Supervisor should receive an email to confirm the CRM ticket has been created with the ticket number 
    - [ ] Email copy should be the default CRM copy when a ticket is created 
- [ ] The ticket should be created 
    - [ ] for Supervisor as the primary contact, not the applicant 
    ~- [ ] with the option "My Team or Organization"~ (no need as this field is hidden in CRM ticket field)
    - [ ] under "Leave & Time Off > Maternity, Parental, Adoption" 

*Note* - Some AC have been moved to 1190 (thread) and 1192(pdf). The estimated efforted was reduced from 8 to 5 as some of the ACs were moved to separate tickets.

**Technical notes**

fazil-ey:
Hey @T854236, will you be able to provide the API documentation for creating CRM tickets

T854236:
@fazil-ey 

Please request engagement for a CRM resource with a JIRA service request.  You'll also need a user license account provisioned to be used by DJ to access the APIs.

MeghanStothers:
Caleb has submitted a JIRA service request

iman-jamali-fw:
@ca-liu 

Hi Caleb. I have a question for you. 

Does CRM have the capability to send an email? We need to send an email once a ticket (incident) is created/updated, wondering if we can leverage CRM REST API to handle that.

MeghanStothers:
Involves the Camunda workflow and CRM integration. 
Estimate based on initial understanding of work. 
Changes required will be put into a separate ticket.

iman-jamali-fw:
Progress report:
- Created a CRM ticket leveraging API
- Added text to that ticket as a thread
- Have a contact assigned to that ticket
- Updating a ticket is not working yet, it seems like an issue from CRM not allowing me to update (Patch) an existing CRM ticket.

Next:
- Learning how to attach a file (PDF) to an incident (ticket). 
- Will meet with @fazil-ey to find out about the required fields and how to map them to the incident object in the API. After that meeting, I'll have a number of questions for the CRM team to ask.

Stella-Archer:
Iman digging into figuring out updating the API
questions for CRM ready for tomorrow (Friday 9 June)

[...]

PR #1220, opened 2023-06-23
feature/DGJ_1100-crm-service-create-incident

iman-jamali-fw:
## Summary

This PR adds a CRM service in Camunda to create a CRM incident (ticket) leveraging its API. #1100 

## Changes
- CRM service is added
- Number Java objects were added for CRM operation

CRM service can be added as a service task in the Camunda workflow to create a CRM ticket for users who run that workflow. 

Screenshot 2023-06-23 at 11 46 46 AM

## Notes
At this point, the service only supports incident creation and saves it to the form. The next steps will involve the ability to update an existing ticket, extracting relevant field in the form and adding them as a thread of text into an incident, and making sure the pdf style and format is according to requirement. These will handle in separate tickets in the future.

### DIFF
--- a/deployment/openshift/bpm_dc.yaml
+++ b/deployment/openshift/bpm_dc.yaml
@@ -249,6 +249,16 @@ objects:
                     configMapKeyRef:
                       name: forms-flow-ai-config
                       key: BPM_CLIENT_CONN_TIMEOUT
+                - name: CRM_URL
+                  valueFrom:
+                    secretKeyRef:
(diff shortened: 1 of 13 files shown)


bhumin-fw (approved):
Great job Iman! This looks good.

One comment, 
"test.pdf" let's change it to "attachment.pdf" (It seems real world word then test) for now if you want to test this on DEV. 
In next update let's fetch PDF name from workflow (I'm sure that's on your radar, just adding it here.)

PR #1234, opened 2023-06-30
DGJ-822 initial setup for test suits integration on Pull request

bhumin-fw:
## Summary
Adding On new PR test suites triggers for below repo
- forms-flow-web
- forms-flow-webapi
- forms-flow-bpm
- forms-flow-documents

Test suits are not setup on below repo. It breaks with SQLAlchemy version <2. In new release new test case added added and version upgrade to 2.017.
- forms-flow-data-analysis-api

## Changes
Added 4 workflow fine that setup service and run test case.
**forms-flow-bpm**
- No Change in test suites

**forms-flow-documents**
- No changes on test suits

**forms-flow-webapi**
- Remove commented code in test suits, It affect on 2 files. This code has been commented for digital journey requirements.

**forms-flow-web**
- Skip 2 file in test suites "baserouting" and "navbar". It fails due to jest configuration with package lock and new install version. Need to handle it separately after latest upgrade.
- Comment out html component test as those are hidden in task page for Digital Journey's requirements.
- Added new attribute in const file base on which test case are running. Also skip one function that not match with output on digital journey requirements.

## Notes
- Keep "develop" branch only. So it won't trigger right now. Need to test first before applying on "main" branch.
- Skip 2 files (navigation and route file only), After upgrade if we are able to fix package-lock.json file then we will re-enable it and changes test case based on updated code and component.
- After upgrade, We needs to fix package-lock.json file and apply the same in testcase setup. Right now It runs by "npm i --legacy-peer-deps"


forms-flow-api/tests/unit/api/test_draft.py
```diff
@@ -129,11 +129,6 @@ def test_draft_submission_resource(app, client, session, jwt):
     assert response.json.get("formId") == payload["formId"]
     assert response.json.get("submissionId") == payload["submissionId"]
 
-    draft = Draft().query.get(draft_id)
-    assert draft.status == "0"
-    assert draft.data == {}
-    assert draft.application_id == response.json.get("id")
-
```
iman-jamali-fw, line 136: I would comment out instead of removing them to make the future upgrade and merge conflict easier.

iman-jamali-fw (approved):
Looks great! 
Just had a suggestion on commenting out the original code.

forms-flow-api/tests/unit/api/test_draft.py
bhumin-fw: Commented code

bhumin-fw:
> Looks great! Just had a suggestion on commenting out the original code.

Comment out no processed test cases.

PR #1239, opened 2023-07-08
DGJ-1198 Hide unassigned task to user if the user is manager

bhumin-fw:
## Summary
DGJ-1198 Hide task that are not assigned to user.

## Changes
- Apply filter assignee if user has role manager
- Hide "assignee" filter option from the search bar.

## Notes
NA

### DIFF
--- a/forms-flow-web/src/services/UserService.js
+++ b/forms-flow-web/src/services/UserService.js
@@ -148,11 +148,13 @@ const getFormioToken = () => localStorage.getItem("formioToken");
 // const doLogin = KeycloakData.login;
 // const doLogout = KeycloakData.logout;
 const getToken = () => KeycloakData?.token;
+const getUserInfo = () => KeycloakData?.userInfo;
 
 const UserService = {
   initKeycloak,
   userLogout,
   getToken,
+  getUserInfo,
   getFormioToken,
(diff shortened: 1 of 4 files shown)


forms-flow-web/src/services/UserService.js
```diff
@@ -148,11 +148,13 @@ const getFormioToken = () => localStorage.getItem("formioToken");
 // const doLogin = KeycloakData.login;
 // const doLogout = KeycloakData.logout;
 const getToken = () => KeycloakData?.token;
+const getUserInfo = () => KeycloakData?.userInfo;
 
 const UserService = {
   initKeycloak,
   userLogout,
   getToken,
+  getUserInfo,
```
iman-jamali-fw, line 157: Do you mind moving `getUserInfo ` to the last item (after `setKeycloakJson`)? It would make the upgrade and merge conflict easier for the next update.

iman-jamali-fw (approved):
Looks great! A remedy to many headaches :)  Thank you Bhumin

forms-flow-web/src/services/UserService.js
bhumin-fw: That make sense.

PR #1240, opened 2023-07-11
DGJ-1131 replace single quote with double single quote while sending data to ODS

bhumin-fw:
## Summary

DGJ-1131 replace single quote with double single quote while sending data to ODS
- Setup dynamic character replace in string value

## Changes
- Add provision in `sendDataToODS` delegates which replace provides characters.
- Added readme file on how to use input param to play with request object to ODS.

## Notes
NA


forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/SendSubmissionToODSDelegate.java
```diff
@@ -87,11 +90,15 @@ private void sendSubmissionToODS(DelegateExecution execution) throws IOException
...
-            json = objectMapper.writeValueAsString(values);
+            values = convertToSnake(values);
+        }
+        if (replaceText != null) {
+            String[][] replaceArray = parseStringToArray(replaceText.split("\\s*,\\s*"));
+            if (replaceArray.length > 0) {
+                values = replaceTextAll(values, replaceArray);
+            }
         }
-        System.out.println(json);
```
iman-jamali-fw, line 94: This line proved to be very useful for debugging purposes. I'd leave it here.

forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/SendSubmissionToODSDelegate.java
```diff
@@ -129,4 +136,39 @@ public Map<String, Object> convertToSnake(Map<String, Object> values) {
...
+        for (int i = 0; i < input.length; i += 2) {
+            if (i+1 <= input.length && init < replaceArraylength) {
+                nestedArray[init][0] = input[i];
+                nestedArray[init][1] = input[i+1];
+                init++;
+            }
+        }
+        return nestedArray;
+    }
 }
```
iman-jamali-fw, line 174: I'm wondering if we change the input type in the workflow from string to an array of strings like ["@, #", "-,+"] makes things easier for the form designer. However, I'm unsure how hard it is to escape ' and " in that case.

forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/SendSubmissionToODSDelegate.java
bhumin-fw: very next line is that prints json value anyway if you pass debug true. That's why remove additional log here.
```
if(debug) {
    System.out.println("Sending values to ODS: " + json);
}
```

iman-jamali-fw: Oh, you're right! I thought that on is the one. No worries then

forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/SendSubmissionToODSDelegate.java
bhumin-fw: @iman-jamali-fw I have moved from string array to list. apply your suggestion it and pushed new code.

iman-jamali-fw: Looks awsome!

iman-jamali-fw (approved):
Looks great!

PR #1248, opened 2023-07-21
DGJ-617 cache employee demographic data

bhumin-fw:
## Summary
DGJ-617 cache employee demographic data

## Changes
- Add API based cache for get ministry list
- Add key based cache for user

## Notes
NA

### DIFF
--- a/forms-flow-api/src/formsflow_api/resources/influenzaWorksite.py
+++ b/forms-flow-api/src/formsflow_api/resources/influenzaWorksite.py
@@ -7,6 +7,7 @@ from formsflow_api_utils.utils import (
     auth,
     cors_preflight,
     profiletime,
+    cache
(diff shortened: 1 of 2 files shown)


iman-jamali-fw (commented):
I would increase the cache timeout to one day as confirmed by the ODS team that changes to ODS are deployed daily.

iman-jamali-fw:
> I would increase the cache timeout to one day as confirmed by the ODS team that changes to ODS are deployed daily.

On second thought, maybe less than one day (e.g., every 12 hours) as the user access time might not be synced with the ODS change window.

iman-jamali-fw (approved):
Great work Bhumin. This brings a lot of value. We should cache as much as possible if this goes well.

PR #1250, opened 2023-07-24
fix/DGJ_1100-handle-user-idir

iman-jamali-fw:
## Summary

A minor PR to modify the CRM client search process in the service by excluding the `@idir` postfix. #1100 

## Changes
- Improved the crmDelegate to accommodate users with the `@idir` postfix in their IDIR usernames.

## Notes
The `@idir` postfix is added to all users in all environments except for the local environment, which has a different setup. The lower Keycloak version in the local environment lacks the idirMapper present in other environments.


eyds-canada-recruitment (approved):
This looks good Iman.

bhumin-fw (approved):
This looks good Iman.

PR #1267, opened 2023-07-29
DGJ_1100-crm-creation-only-on-supervisor-validate

iman-jamali-fw:
## Summary

This PR ensures that CRM creation happens only when the supervisor validates the submission, also skip the CRM creation if it already been created. #1100 

## Changes
- Adding a conditional to the CRM service to skip the CRM operation if `crmId` already exists.
- Adding a conditional to the mat/pat workflow to only run CRM operation if the supervisor validates the submission.
- CRM product and category were added to the CRM incident. (this is to ensure tickets are categorized under `Leave & Time Off > Maternity, Parental, Adoption`

Part of the mat/pat workflow that was updated:
Screenshot 2023-07-28 at 6 40 38 PM

## Notes
Some of the logic needs to be improved once the CRM update ticket is implemented.


bhumin-fw (approved):
Looks good Iman!

PR #1282, opened 2023-08-16
DGJ-1190 Mat-Pat CRM ticket thread integration

bhumin-fw:
## Summary
DGJ-1190 Added Mat-Pat form fields in CRM ticket thread

## Changes
- Added Channel and content type to handle CRM thread configuration
- Added thread with HTML/CSS channel and content type as HTML/TEXT
- Thread text will be calculate on form itself with hidden field `threadText`
- `threadText` should be pass as input param in workflow when invoking crmDelegate.

## Notes
NA

### DIFF
--- a/forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/CrmDelegate.java
+++ b/forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/CrmDelegate.java
@@ -93,7 +93,6 @@ public class CrmDelegate extends BaseListener implements JavaDelegate {
         Map<String, String> ids = extractIds(formUrl);
         String formId = ids.get(FORM_ID);
         String submissionId = ids.get(SUBMISSION_ID);
-        String threadText = String.valueOf(execution.getVariables().get(CRM_THREAD_TEXT));
         
         if (formId == null || submissionId == null) {
             System.out.println("formId or submissionId is null. formUrl: " + formUrl + " formId: " + formId + " submissionId" + submissionId);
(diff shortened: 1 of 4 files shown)


iman-jamali-fw:
Looks great! especially with html type, it's super flexible. 

I wonder about the requirement of `threadText` in the workflow as input param, why it's needed?

forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/CrmDelegate.java
```diff
@@ -117,7 +121,7 @@ private void crmOperation(DelegateExecution execution) {
         }
 
         // Create a new incident in CRM
-        CrmPostResponse crmPostResponse = createCrmIncident(contactId);
+        CrmPostResponse crmPostResponse = createCrmIncident(contactId, threadText);
```
iman-jamali-fw, line 124: What do you think of passing the `execution` object to the `createCrmIncident` method and then load form properties in that method? We'll need to load several properties in that method including `subject` which I'm currently working on. This way we won't need to pass several parameters to `createCrmIncident`, the method responsible for loading relying on `execution` object.

bhumin-fw: I was suppose to discuss the same with you. As Even in thread we have possible 3 params that we might start using in future let me fix that.

bhumin-fw:
> I wonder about the requirement of `threadText` in the workflow as input param, why it's needed?

It is not required. We can add hidden field with name "threadText" then it will work without passing params.

iman-jamali-fw (approved):

PR #1284, opened 2023-08-17
DGJ_1224-CRM-ticket-field-mapping

iman-jamali-fw:
## Summary

This PR adds two fields to the CRM tickets, subject, and assignee.

## Changes
- CRM subject was updated to include the submitter's name on it.
- CRM assignee field was added.


forms-flow-bpm/src/main/java/org/camunda/bpm/extension/hooks/listeners/CrmDelegate.java
```diff
@@ -62,6 +64,9 @@ public class CrmDelegate extends BaseListener implements JavaDelegate {
     private static final String CRM_LOOKUP_NAME = "crmLookupName";
     private static final String CRM_MAT_PAT_PRODUCT_LOOKUP_NAME = "Leave & Time off";
     private static final String CRM_MAT_PAT_CATEGORY_LOOKUP_NAME = "Maternity, Parental and Adoption Leave";
+    private static final String CRM_MAT_PAT_STAFF_GROUP_LOOKUP_NAME = "TES HR Admin Technicians";
+    private static final String CRM_MAT_PAT_SUBJECT = "Maternity and parental leave form";
```
bhumin-fw, line 68: Can we pull this from execution.getVariables? we need to pull form name. So it will stay dynamic and we do not need to make any change in code if we change anything in form name.

PR #1287, opened 2023-08-24
DGJ-1279 SL admin group with view all submission rights

bhumin-fw:
## Summary

DGJ-1279 Allow group of SL review admin to view and print submissions

## Changes
- Added code that can extend for view only on WEB
- Added sl-review-admin group and key to allow user to fetch application list
- Added formio roles for sl-review-admin in utils.
- Added sl-review-admin in formio formflow-template.json file

## Notes
NA


iman-jamali-fw (approved):
It looks good to me. 
Great work Bhumin!

PR #1293, opened 2023-08-25
DGJ-1291 in edit form, Print to PDF is trim text and show only 1st line

bhumin-fw:
## Summary

DGJ-1291 in edit form, Print to PDF is trim text and show only 1st line

## Changes
- Select all text area field and hide at the time of print.
- Pull text from text area and add in DIV
- Remove temp divs for text and Show textarea again.

## Notes
NA


iman-jamali-fw (approved):
It looks good to me

PR #1295, opened 2023-08-25
DGJ_1224-crm-priority-duedate-field

iman-jamali-fw:
## Summary

This PR adds a new field to the CRM ticket, `priority duedate that will be extracted from the form.

## Changes
- Add a new property to the CRM ticket, loaded form the form called `priorityDuedate`
- Unrelated to this PR, email logger was disabled to not to log email data when a DMN workflow is running

## Notes
I'll add the logic to capture that date based on the mat leave date.


bhumin-fw (approved):
Looks good Iman!

Issue #895, opened 2023-02-08
Mat/Pat : Supervisor Approval


fazil-ey:
**User Story**
As a supervisor I should be able to approve or return forms submitted by employees under me so that their leave forms can be sent for further processing. 

**Note** - need to confirm the status so that we are consistent across DJ. To be discussed when this ticket is picked up. 

**Acceptance Criteria**

- [x] Supervisor is able to see all forms submitted by applicants who specified them as their supervisor in the form
     - Status of the form on submission (or resubmission) is "NEW" 
- [x] Supervisor is able to see the following in the task name 
      - Name of the form with applicants name "Maternity, Parental Leave and Allowance form for *applicants name*"
      - Status of the application i.e "NEW" or "APPROVED" or "DENIED" 
- [x] Supervisor is only able to view the review page and certification page of the form. All other tabs should be hidden (except those needed technically)
- [x] Supervisor is able to approve the form. 
     - Status of the form on approval is "APPROVED" 
- [x] Supervisor is able to send back the form for resubmission if any changes are required
     - Status of the form on sending back is "DENIED"

MeghanStothers:
Re-work the tickets with consistent naming conventions to see if we need new ones

iman-jamali-fw:
@fazil-ey 

These are the status we currently have in different forms.
- Telework: `New`, `Approved`, `Denied`
- SL: `Submitted - Validation`, `Submitted - Final`
- Cold/flu: `Submitted`
- Other status in the system (used at some point): `Resubmit`/`Resubmitted`, `Submitted`, `validated`

iman-jamali-fw:
progress report:
-  [x] The new status for the Mat/Pat form was added which no longer relies on workflow and its version, which can be changed anytime by the form designer
- [x] Updated to correctly assign edit access to statuses used in multiple forms.
- [ ] Pending: Display the applicant's name and submission status under the "Review and Process Forms" tab.

Expect to have this ticket ready for PR today

iman-jamali-fw:
@fazil-ey 
Let me know if `status` looks good on the task item. It'll be like: Maternity and parental leave form for {applicants name} **- {status}**

example below for the status of NEW:

[...]

PR #1354, opened 2023-09-25
DGJ_895-mat_pat-new-status

iman-jamali-fw:
## Summary

This PR adds a new form edit access system that allows each form to control which status should be allowed for user access. #895 

## Changes
- Replacing the list of status that supports editing, with a new form-mapper that allow each form to control its status for edit submission access
- Updated the components that deal with edit submission to work with the new access system.
- Added new status to the Mat/Pat form: (`NEW`, `APPROVED` and `DENIED`)
- Apart from this PR, the mat/pat form was updated to support the new status and its workflow now has `status` as part of the task name displayed in the list under the `/task` page.

### DIFF
--- a/forms-flow-web/src/components/Application/List.js
+++ b/forms-flow-web/src/components/Application/List.js
@@ -27,7 +27,7 @@ import {
   MULTITENANCY_ENABLED,
   STAFF_REVIEWER,
 } from "../../constants/constants";
-import { CLIENT_EDIT_STATUS } from "../../constants/applicationConstants";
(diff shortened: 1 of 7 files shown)


bhumin-fw (approved):
Great update Iman! looks more controlled. 
let's document that as well, if form name or status change we needs to edit this code for the same.

iman-jamali-fw:
> Great update Iman! looks more controlled. let's document that as well, if form name or status change we needs to edit this code for the same.

Sure, will add documentation later today.

PR #1399, opened 2023-10-24
DGJ-1324 Update status when CRM update trigger

bhumin-fw:
## Summary
DGJ-1324 Update status when CRM update trigger

## Changes
Added default status to "Unresolved" and when ticket is update new status set as "Updated".

## Notes
NA


iman-jamali-fw (approved):
Looks good to me.

Issue #1302, opened 2023-09-06
Mat/Pat: Server-side pdf generation timeout issue


iman-jamali-fw:
Random Mat/Pat submissions fail with a timeout error(504). The issue is thrown at the PDF-generation level. Adding PDF generation resources seemed to reduce the likelihood of the error thrown but hasn't fixed the root cause of the issue.

AC: Fixing the PDF generation timeout issue on Mat/Pat submissions.

Technical details: There are multiple timeout variables defined in Camunda that might have something to do with this issue. Finding and increasing those timeout variables might help fix this problem.

Workaround - Usuaully happens when supervisor approves. workaround is to refresh the page and validate again

iman-jamali-fw:
Progress Report:

**Learning So Far:**
I've learned that the Camunda workflows for Mat/Pat always complete successfully, but there's an issue with occasional delays in starting, leading to a 504 timeout error when the website (Review page) communicates with Camunda.

**What I've Tried:**
I've attempted the following:
1. Modified the Mat/Pat BPMN workflow, but it didn't solve the problem.
2. Increased Camunda's memory and CPU resources, which also didn't work.
3. Scaled the number of Camunda pods on OpenShift without success.
4. Tried extending the website's HTTP call timeout using NGINX, which didn't resolve the issue.

**What I'll Do Next:**
My next steps:
1. Explore and adjust various Camunda timeout settings, as there are multiple configurations.
2. Increase the React app (website) HTTP client timeout to accommodate potential delays in network calls.

I'm working to better understand and resolve the intermittent 504 timeout error.

iman-jamali-fw:
## Cause of the timeout issue
The problem stemmed from the short connection timeout of the Camunda DNS server (Route). The default timeout in Openshift Route is set at 30 seconds, which proved insufficient for processing Mat/Pat workflows. These workflows on the DEV environment could take up to 2 or 3 minutes.

## Resolution
To address the timeout issue, all timeout-related configurations in various components at different layers were extended to accommodate longer Camunda operations. Ensuring longer timeouts required adjusting the limits at all layers of connection for each engagement component in the workflow. This involved modifying timeout settings for:

- DNS server timeout (Routes in Openshift)
- Gateway servers, such as NGINX for the Website, GUNICORN for API and PDF generation, and the embedded Spring Boot server for Camunda.

[...]

PR #1404, opened 2023-10-25
DGJ_1302-camunda-timout-gatway-504-issue-fix

iman-jamali-fw:
## Summary

This PR set the HTTP call timeout limit for all relevant components involved in the platform to 300 seconds. Make sure the current long workflow (mat/pat) processes can go through and guarantee future calls have enough time to send a response to the server.

The reason you see all timeout configs were changed from 600s to 300s is that I've been pushing changes directly to main for troubleshooting and now that identity the fix, making the change a more reasonable value (i.e., 300s).

Please read more about the details of the issue and fix it at #1302


deployment/openshift/bpm_dc.yaml
```diff
@@ -418,7 +418,7 @@ parameters:
   - name: CAMUNDA_JOB_MAX_WAIT
     description: CAMUNDA_JOB_MAX_WAIT
     required: true
-    value: "60000"
+    value: "300000"
```
bhumin-fw, line 421: Iman can you review this specifc value, Seems we don't want to increase waiting time from 1min to 5min. If this cause any and increasing time required then continue with 5min for now.

bhumin-fw (approved):
Looks good to me! great job Iman to trace down and fix this long process issue.

Issue #1349, opened 2023-09-21
Tool - Explore wizard Form issues that dont work, but work for regular forms


fazil-ey:
There are some bugs in the wizard forms that work for normal forms. e.g. links opening in the same tab, screen not scrolling up to the error on the page. We need to explore how this works in wizard forms

 
Related to #1342,  #1274

- [x] Explore why links are not opening in a new tab
- [x] Explore why page does not scroll to the top when there is an error on the page

iman-jamali-fw:
Progress report:
Added a PR for fixing openning links in separate tabs. Will continue working on fixing scroll to validation errors today.

iman-jamali-fw:
First AC: Explore why links are not opening in a new tab was done and ready for testing in #1274 ticket.

iman-jamali-fw:
Second AC: **_Explore why page does not scroll to the top when there is an error on the page_** was done and ready for testing in #1342 ticket,

PR #1433, opened 2023-11-01
DGJ_1349-open-links-in-new-tabs

iman-jamali-fw:
## Summary
This PR (#1349) improves the existing function that converts all the form's links to open in new tabs. Instead of modifying the live form after it's been rendered, it changes the links before the form is rendered (changes the static form). This will ensure all the links even those that are conditionally rendered in the line form (used to miss the _blank attribute) are opening in new tabs.

One improvement over my approach would be to modify the form right after/before added to the state. I wasn't able to find the right place where Redux adds form to the app state and instead changes the form state in each component (e.g., view, draft, edit submission etc). Please let me know if you know how to change the form there so we modify the form in one place instead of all view/edit/draft files.

### DIFF
--- a/forms-flow-web/src/components/Draft/Edit.js
+++ b/forms-flow-web/src/components/Draft/Edit.js
@@ -65,7 +65,6 @@ const View = React.memo((props) => {
     (state) => state.formDelete.formSubmitted
   );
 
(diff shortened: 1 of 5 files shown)


bhumin-fw (approved):
LGTM!

PR #1438, opened 2023-11-09
DGJ_1425-camunda-crm-connectio-issue-add-logs

iman-jamali-fw:
## Summary

Added logs to learn more about Camunda-CRM connection issue. Also added HTTP error handler to CRM connections.


bhumin-fw (approved):
Looks good Iman

PR #1441, opened 2023-11-10
DGJ_1425-camunda-crm-connectio-issue-change-http-client

iman-jamali-fw:
## Summary

CRM HTTP client library was replaced with the Java standard one for debugging. Hopefully, this helps reveal the connection timeout mystery 🤞


jerry-ey (approved):
The syntax looks good to me, although I am not familiar with JAVA.

PR #1468, opened 2023-11-23
DGJ-1427 | submission loader in Supervisor view

jerry-ey:
## Summary

Add loading spinner when hit "validate submit" button in Supervisor page.

## Changes

- Add event handler to `ServiceTaskDetail.js`, it passes customEvent props to `<Edit />`, that's why it override the existing event handler in the `<Edit />`
- No need to change in formio.

image


bhumin-fw (approved):
LGTM

PR #1474, opened 2023-11-24
DGJ-1381 send an email to supervisor when CRM ticket update

bhumin-fw:
## Summary

DGJ-1381 send an email to supervisor when CRM ticket update

## Changes
- Call `/incidentReponse` endpoint to update CRM incident with send email configuration
- Keeping existing update flow as it is, because `/incidentReponse` does not support status and contact param

## Notes
NA


jerry-ey (approved):
LGTM!

iman-jamali-fw (approved):
Looks good to me!

PR #1476, opened 2023-11-24
DGJ-1357 | remove workflow name from all forms

jerry-ey:
## Summary

Got final confirmation that we remove workflow name from all forms.


bhumin-fw (approved):

iman-jamali-fw (approved):
Looks good!

PR #1480, opened 2023-11-28
Updated action file for web deployment

kiran-aot:
## Summary
Encountered a deployment issue during the Git Action for forms-flow-web from the helm-web directory, with the error message: "error: error parsing STDIN: error converting YAML to JSON: yaml: line 2: mapping values are not allowed in this context.

## Changes

- Updated helm installation command in forms-flow-web-cd.yml file

image

After updating the helm installation command

image


kkapoor-fresh (commented):
Just wanted to understand the change that we are passing the name "helm-web" twice in the beginning rather than passing the folder ./helm-web at the end of the command and also removed the piping in the command.

kiran-aot:
While we are using helm we need to specify the release name and chart directory , so here I added both as "helm-web", can change to "forms-flow-web". And we don't need to specify files in the directory using oc apply command since helm will catch all the files in chart directory

PR #1491, opened 2023-12-01
DGJ-1488 Skip delete submission and application if it's SL review form

bhumin-fw:
## Summary

DGJ-1488 Skip SL review delete operation

## Changes
- added condition that skip SL review form in delete operation

## Notes
NA


jerry-ey (approved):
LG!

PR #1506, opened 2023-12-07
DGJ-1427 | add option param to hide the 3 dots loader in customEvent

jerry-ey:
## Summary
image

Removing 3 dots loader and only keep the "Submitting loader".


forms-flow-web/src/components/ServiceFlow/details/ServiceTaskDetails.js
```diff
@@ -242,7 +242,8 @@ const ServiceFlowTaskDetails = React.memo(() => {
 
   const onFormSubmitCallback = (actionType = "", successPage) => {
     if (bpmTaskId) {
-      dispatch(setBPMTaskDetailLoader(true));
+      // comment out the following line. It is showing the 3 dots animation.
+      // dispatch(setBPMTaskDetailLoader(true));
```
iman-jamali-fw, line 246: @jerry-ey 
If we disable the default loaded, does it mean all the forms should send an event to trigger the loader onSubmission?

Can you please test other forms with this change to see their current behaviour doesn't break? I belive only Telework form send event for loader.

jerry-ey: Good call, I will double check and make necessary changes.

jerry-ey:
Some forms don't have customEvent, so in order not breaking other form, this change will only apply to Mat/Pat with new param added to customEvent.

Will add this `isDefaultLoaderHidden: true` to DEV and TEST. Please review again.

image

iman-jamali-fw (approved):
Looks great Jerry!
Please just document the process before merging. Thanks

jerry-ey:
> Looks great Jerry! Please just document the process before merging. Thanks

No problem. Will add the detail.

jerry-ey:
Added document.

bhumin-fw (approved):
Looks good!

PR #1540, opened 2024-01-05
DGJ-1453 | fix the wizard container css to align better buttons position

jerry-ey:
## Summary

Align the buttons to look niser.

## Changes

- Fix the `.formio-wizard-nav-container ` css.
- The issue is because it has weird left and bottom margin, so set to `margin: 0`
- Use `gap: 5px` to replace each button `margin-left: 5px` to fix if there is only one button in the flex container.

### Supervisor View
image

image

### Applicant View
image


iman-jamali-fw (approved):
LGTM!

bhumin-fw (approved):
Looks good to me. Please check with all the forms.

PR #1588, opened 2024-02-01
DGJ-1501 | disable console.warn in the function

jerry-ey:
## Summary

- Remove unnecessary console.warn
- After digging the code for a while, `console.warn` is happening in formio source code `getForm`, so nothing we can do but put the `console.warn` to be empty function in that scope, not effecting the global scope. I believe this is the best way for this case.

https://github.com/formio/react/blob/master/src/modules/form/actions.js

image


jerry-ey:
- in the Form page, showing console.warn
CleanShot 2024-02-02 at 13 33 03@2x

- in Review Page, disable console.warn
CleanShot 2024-02-02 at 13 34 19@2x

bhumin-fw (approved):
Looks good! Jerry

PR #1600, opened 2024-02-01
Remove margin for hidden alert components

Andrew-Vargas-bcgov:
Removes margin for hidden alert components so that they do not take create excess whitespace.


jerry-ey (approved):
LG!

bhumin-fw (approved):

PR #1601, opened 2024-02-01
DGJ-1521 Return count of job records for logged in employee

bhumin-fw:
## Summary
DGJ-1521 Identify when user have more than 1 jobs.

## Changes
return count of jobs when fetching employee data from employee_demo endpoint from ODS>

## Notes
NA


jerry-ey (approved):
LGTM!

PR #1615, opened 2024-02-08
DGJ-1612 Filter only assigned task for formflow-client role

bhumin-fw:
## Summary

DGJ-1612 Filter only assigned task for formflow-client role

## Changes
 - Added CLIENT role in filter for assignee user name on service task list API call.

NA

## Notes
NA


jerry-ey (approved):
LGTM!

PR #1617, opened 2024-02-09
DGJ-1597 | search submitDisplayName in submitted page

jerry-ey:
## Summary

DGJ-1597 | search submitDisplayName in submitted page

## Changes

- add new field in db under application table
- when frontend send fetch request with param `applicationName`, we also search `submission_display_name`

There are 2 scenario that  `submission_display_name` will be created under application table
1. when draft was created
2. when user submit the form

### Testing
I have tested 2 forms
- Telework has no `submission_display_name` hidden field, no error, but not able to search this 
- MatPar form has `submission_display_name` hidden field, working fine.

1. search name `jer`
CleanShot 2024-02-08 at 19 39 11@2x

4. without any search
CleanShot 2024-02-08 at 19 39 36@2x


forms-flow-api/src/formsflow_api/models/application.py
```diff
@@ -201,7 +206,11 @@ def filter_conditions(cls, **filters):
             FormProcessMapper.process_name.label("process_name"),
             FormProcessMapper.process_tenant.label("process_tenant"),
         )
-        query = query.filter(*filter_conditions) if filter_conditions else query
+        if filters["application_name"]:
+            filter_conditions.append(Application.submission_display_name.ilike(f'%{filters["application_name"]}%'))
+            query = query.filter(or_(*filter_conditions)) if filter_conditions else query
```
jerry-ey, line 211: key part

forms-flow-api/src/formsflow_api/schemas/application.py
```diff
@@ -63,8 +63,8 @@ class Meta:  # pylint: disable=too-few-public-methods
     submission_id = fields.Str(data_key="submissionId")
     form_url = fields.Str(data_key="formUrl", load_only=True)
     web_form_url = fields.Str(data_key="webFormUrl", load_only=True)
-
-
+    data = fields.Dict(data_key="data", load_only=True)
```
bhumin-fw, line 66: Why we need to load "data"?

forms-flow-api/src/formsflow_api/models/application.py
bhumin-fw: Seems fine, Added comment before or inside if condition about change.

bhumin-fw (approved):
Looks good! 
In draft.py condition like seems longer then X characters, it will fail in lint. Double check it and break down it to align with lint.

PR #1619, opened 2024-02-12
DGJ-1613 | Cold/Flu | Success page | add dynamic year instead of hardcoded

jerry-ey:
## Summary

- Success page from 2023 to 2024
- Make it more dynamically


bhumin-fw (approved):
Looks good!

PR #1661, opened 2024-03-05
DGJ-1652 | Cold/Flu | Admin user can edit on "submitted" page.

jerry-ey:
- Fix the constant text and then it will show "Edit" button on submitted page.
- This line did include cold-flu-admin to show all applications on submitted page, so we don't need to add more.
CleanShot 2024-03-05 at 07 59 16@2x


bhumin-fw (approved):
Looks good.

PR #1702, opened 2024-04-08
DGJ-1633 | update release note

jerry-ey:
## update release note content

- This json is what I used for via postman in TEST env. 
-  And we will use this json file to send post request again when we deploy to PRD.
- I already updated the release note in database in Test env.

CleanShot 2024-04-08 at 11 44 00@2x


bhumin-fw (approved):
Looks good!

PR #1703, opened 2024-04-08
DGJ-1550 Added success page for Delete telework confirmation

bhumin-fw:
## Summary

Added Success message on confirmation page

## Changes
Adde one more condition in success page.

Screenshot 2024-04-08 at 1 00 11 PM

## Notes
NA


forms-flow-web/src/components/Form/Item/Submission/Success/SuccessPage.js
```diff
@@ -646,6 +646,24 @@ export default React.memo(() => {
           </div>
         </>
       );
+    } else if (
+      search.includes(successTypes.DELETE_TELEWORK)
+    ) {
+      let name = search.replace("?type=DELETE_TELEWORK_", "");
```
jerry-ey, line 652: maybe use `successTypes.DELETE_TELEWORK` here, instead of the string. what do you think?

bhumin-fw: that's good catch, Updated it. Please review.

jerry-ey (approved):
Looks good!

PR #1795, opened 2024-05-27
DGJ-1789 guest login hide draft and application pages

bhumin-fw:
## Summary

Added guest login detail on keycloak.
Hide draft and application page from menu, list and set redirect on /form for guest user.

## Changes
- Added user name **anonymous_guest** in keycloak with role **forms-flow-client** and **anonymous**.
- From the existing const use **anonymous_user** and add conditions on draft, application and form page to hide unnecessary content.

NA

## Notes
NA


jerry-ey (approved):
Looks great!

PR #1815, opened 2024-06-07
DGJ-1808 | Redirect to success page on Draft page after submit 

jerry-ey:
- Redirect to success page on Draft page after submit instead of redirecting to draft page again

CleanShot 2024-06-07 at 09 21 14@2x


bhumin-fw (approved):
Great work, It looks good.

PR #1823, opened 2024-06-13
Dgj 1759 crm category update based on id and not name

bhumin-fw:
## Summary

Select the category based on ID instead of name if provided in the form. Some end categories are similar under different main categories. For example, `Maternity Leave` falls under both "`Maternity, Parental, and Adoption Leave`" and "`T2 Payroll (Transaction)`". In this case, we need the ID of the selected category.

## Changes
- When we provide `crmCategoryLookupId` in the form, it will prepare a POST request for CRM incident with the ID lookup.
- If it is not provided, it will continue using the `crmCategoryLookupName` field to prepare the POST CRM incident request.
- Ignore product and assignTo at time of CRM update request, It is enabled for all forms for now.

## Notes
Using an ID to lookup a category is not advisable and should be used only in certain cases. Along with crmCategoryLookupId, one must provide the unique category name of the required category or its parent category in the crmCategoryLookupName field.


c1495616js (approved):
Good job!

Andrew-Vargas-bcgov (approved):

PR #1834, opened 2024-06-19
Dgj 1819 rename file attachment in crm

bhumin-fw:
## Summary

Rename CRM file attachments

## Changes
Change code to pull new data from the same hidden field, crmFileFields
Now it supports dynamic filename, static filename and prefix.

## Notes
NA


jerry-ey (approved):
LG!

PR #1850, opened 2024-06-28
DGJ-1835 allow multiple file to be attached from same component to email

bhumin-fw:
## Summary

Attach multiple file from base64 and digital-journey source to email

## Changes
Allow base64 file to attach to email also allow multple file from single component to be attached to email.

NA

## Notes
NA


jerry-ey (approved):
Looks Great!

PR #1967, opened 2025-04-28
Add leadership survey submission handling and update constants

Andrew-Vargas-bcgov:
Added success page and form submission access for the new senior leadership feedback survey


bhumin-fw (approved):
We need to update the status configuration as well (which we can discuss later when we know all status). As far as the constants for the success page, everything looks good.

